PR #3, opened 2019-02-15
Add Mixed use of composer and zip modules

eternoendless:


mickaelandrieu:
Wait! looks like the "slowness" of the build doesn't come from this decision as some builds are green again [without any revert](https://travis-ci.org/PrestaShop/PrestaShop/builds/493847825)!

I'd like to have more information here (like Blackfire profiles on different environments) before we validate this decision please: as I don't see any hurry here as 1.7.6 is planned for June.

0002-mixed-use-of-composer-and-zip-modules.md
```diff
@@ -0,0 +1,39 @@
...
+In difference with the normal behavior described above, a module installed using composer will have its autoloader and dependencies merged into the core's and placed in the core's `vendor` directory.
+
+This can prove problematic for native modules:
+
+- When the module is uninstalled, both its dependencies and autoloader are left behind in the core.
+- If the module is upgraded using the marketplace sources, the dependencies are now available twice: once in the core and once in the module.
+
+Alternate systems to avoid having modules leak code into the core have been proved unpractical:
+
+- Including modules via [Composer script handler](https://github.com/PrestaShop/composer-script-handler) or git clone is too slow.
```
mickaelandrieu: Looks like it works well => https://travis-ci.org/PrestaShop/PrestaShop/builds/493847825 and the slowness we have encountered these days was not related to the extension.

The impact on time of the build is ~ 40%: while it's a lot, it's not what we encounter these days when @PierreRambaud says its GCP build encounters a difference from 1'30m to 15 minutes! (~ +1500% !)

mickaelandrieu: while I agree it's still "slow", it's not that slow for a build operation (and we can invest time to improve the performance of the process while I've already suggested to @PierreRambaud)

eternoendless: I experienced the slowness firsthand unfortunately.

PR #7, opened 2020-02-13
0006 - Module as a Bundle

sowbiba:


0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,66 @@
...
+
+Bundles are registered when the Kernel boots and each bundle can register its own services, routing, controllers, templates, ... and this is automated by Symfony when the structure conventions are respected.
+
+When registering a bundle, you have to instanciate the main class which extends `Symfony\Component\HttpKernel\Bundle\Bundle` and add it to the returned array of method `registerBundles` of the AppKernel. And that's it.
+
+But, how can we know the full class name of a Module's bundle when
+
+
+* Bundle class can be theorically in any sub-directory of the module.
+* We don't have conventions about the namespace : A contributor company can prefix the namespace by it's organization name.
```
eternoendless:
```suggestion
* There are no conventions about what a module namespace should look like.
```

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,66 @@
...
+
+https://symfony.com/doc/3.4/bundles.html
+
+https://symfony.com/doc/3.4/bundles/best_practices.html
+
+Bundles are registered when the Kernel boots and each bundle can register its own services, routing, controllers, templates, ... and this is automated by Symfony when the structure conventions are respected.
+
+When registering a bundle, you have to instanciate the main class which extends `Symfony\Component\HttpKernel\Bundle\Bundle` and add it to the returned array of method `registerBundles` of the AppKernel. And that's it.
+
+But, how can we know the full class name of a Module's bundle when
```
eternoendless:
```suggestion
But, how can we know the fully qualified class name of a Module's bundle when:
```

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+* Bundle class can be theorically in any sub-directory of the module.
+* We don't have conventions about the namespace : A contributor company can prefix the namespace by it's organization name.
+
+Other aspects have to be taken in consideration :
```
eternoendless:
```suggestion
Other aspects have to be taken in consideration:
```

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,66 @@
...
+
+* We will have bundled and not-bundled modules
```
eternoendless:
```suggestion
* PrestaShop will support bundle-like and non bundle-like modules
```

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,66 @@
...
+* A bundle can maybe have to include third-party bundles (?)
```
eternoendless:
```suggestion
* A bundle may need to include third-party bundles (?)
```

Not sure about the need for a module to include or initialize a third party bundle. Do we really need to account for that at this stage? ping @jolelievre

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,66 @@
...
+* A module can be in the filesystem but not activated
```
eternoendless:
```suggestion
* A module can be present in the filesystem but not be activated
```

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Decision
+
+1. Each module which have to be treated like a bundle must have a `Manifest.php` file at the root of it with the list of bundles to load.
```
eternoendless: Best to use lower case for non-class files.

```suggestion
1. Each bundle which have to be treated like a bundle must have a `manifest.php` file at its root. This file will instantiate the module's bundles, then return a list of instances.
```

The original idea was to have a php file because it would instantiate the bundles on its own. If it doesn't, then we could do with a yml file instead of a php file. But I still think a php file would provide more flexibility in case the bundles would need constructor parameters.

In that case maybe a `bundles.php` would make more sense?

BTW having the file `return` by itself may make it more difficult to include. Also since it's an executable maybe there should be more foolproof. I think it would be better to have it declare a callback:

```php
<?php

$registerBundle = function() {
    return [
        new \SomeVendor\SomeModule\SomeModuleBundle(),
    ]
}
```

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+2. The bundles to load must be in an array and instanciated.
+3. If one of the bundles doesn't extends `Symfony\Component\HttpKernel\Bundle\Bundle` the Manifest will be ignored.
+4. The module's bundles will be aggregated into the file `_PROJECT_DIR_/config/module_bundles.php`.
```
eternoendless: This probably should be created in the `var/cache/_ENV_` directory instead so that it can be cleaned.

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+
+* We will have bundled and not-bundled modules
+* A bundle can maybe have to include third-party bundles (?)
+* A module can be in the filesystem but not activated
+
+## Decision
+
+1. Each module which have to be treated like a bundle must have a `Manifest.php` file at the root of it with the list of bundles to load.
+2. The bundles to load must be in an array and instanciated.
+3. If one of the bundles doesn't extends `Symfony\Component\HttpKernel\Bundle\Bundle` the Manifest will be ignored.
```
eternoendless: I guess this would be done at the AppKernel level

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+<?php
+return [
+	'module_wake_up' => [
+		new MyCompanyName\Module\CoffeMakerBundle(),
+		new ThirdPartyCompany\AdditionalBundle(),
+	],
+	'ps_other_module' => [
+		new Prestashop\Module\OtherModuleBundle(),
+	],
+];
```
eternoendless: I don't think we'll be able to easily generate this file (we could have if manifests were static and not php files).

This could be replaced by a cache file stating which modules are bundle-like to avoid going through all active modules.

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+
+https://symfony.com/doc/3.4/bundles.html
+
+https://symfony.com/doc/3.4/bundles/best_practices.html
+
+Bundles are registered when the Kernel boots and each bundle can register its own services, routing, controllers, templates, ... and this is automated by Symfony when the structure conventions are respected.
+
+When registering a bundle, you have to instanciate the main class which extends `Symfony\Component\HttpKernel\Bundle\Bundle` and add it to the returned array of method `registerBundles` of the AppKernel. And that's it.
+
+But, how can we determinate the full class name of a Module's bundle when
```
eternoendless:
```suggestion
But, how can we determinate the fully qualified class name of a Module's bundle when:
```

0004-module-as-a-bundle-registration.md
sowbiba: cleaning it means regenerate it if the file is not present
I thought we want to avoid browsing modules folder again and again ?

sowbiba:
Without inflence the discussion, here is what I imagine

https://github.com/PrestaShop/PrestaShop/pull/17580/commits/669a1192a7cfb428ca53fb4b730711db1569ac2f

0004-module-as-a-bundle-registration.md
jolelievre: This can be done using `Symfony\Component\Config\ConfigCache` I think it's the perfect candidate for this feature
You can define its resources (files that are used to generate it), and then dump your config Then when the config is used it check if `$cacheConfig->isFresh()`:
- in prod environment as long as the dump file exists it's always fresh
- in dev environment it checks if the file exists and if its resources have been modified

It is used by the symfony framework for its container cache policies, and it's been used in our `PrestaShop\PrestaShop\Adapter\ContainerBuilder` as well to improve performance in dev environment

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+
+Other aspects have to be taken in consideration:
+
+* PrestaShop will support bundle-like and non bundle-like modules.
+* A bundle may need to include third-party bundles (?)
+* A module can be present in the filesystem but not be activated.
+
+## Decision
+
+1. Each module which have to be treated like a bundle must have a `manifest.php` file at the root of it with the list of bundles to load.
```
jolelievre: Since it only contains a bundle list why not name it `bundles.php` ? Unless some other information might be added later in this?

0004-module-as-a-bundle-registration.md
jolelievre: It depends on what is allowed to be added in the manifest files, if it's a plain php array of bundles then I don't think it's an issue However it the manifest file incorporates some logic, like dynamically generate the list of bundles based on some config then indeed it won't be possible

But do we want to authorize this? Any complex behaviour like feature toggling and such should be handled via the Bundle configuration, but not through this manifest file in my opinion

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+		new Prestashop\Module\OtherModuleBundle(),
+	],
+];
+```
+
+Indexing the generated array by the module name allow us to don't load bundles of disabled modules.
+
+## Points of friction
+
+The module developer must be careful about the third-party bundles he use. If a component is already used by the Prestashop Core, the module one will be used (because it is loaded last) and that can lead to unpredicted behaviours. (@TODO test this assert).
```
jolelievre: Yes we need to think about "bundle collisions", and it can happen with core bundles, but also with other modules bundles (which will very likely happen, even more than with the core ones that don't need to be included by the module)

However an important thing: Modules MUST NOT override the core bundle, just like their vendor dependencies don't override the core dependencies (we ensure this via a `"prepend-autoloader": false,` in the composer.json)

Also maybe we should include some checker with this feature, that would be able to detect potential conflicts between modules (at the very least as a warning). Because there might be several types of conflicts:
- same bundle but different versions (which one is picked then)
- same bundle but different configurations (are the two configurations able to coexist? or does one overrides the other?)

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,76 @@
...
+
+In discussion
+
+## Context
+
+The normal behavior of a module is to be self-contained, meaning that its code and all its dependencies are stored in the module's directory.
+
+As we are migrating to Symfony, the necessity to use some techniques like DependencyInjection, EventListener, ... come to us.
+
+To avoid scanning the modules directory everytime to know if such services exists within some of the existing modules, the idea came to consider a module like a bundle.
```
jolelievre: It's not exactly true, even as we speak the services are only scanned once when the container is built (at least in prod environment) The advantage of allowing the use of bundle is more:
- offer a true Symfony environment/configuration for modules
- allow modules to integrate some third party modules/extension into PrestaShop

0004-module-as-a-bundle-registration.md
```diff
@@ -0,0 +1,91 @@
+# 4. Registration of a bundle-like module
```
eternoendless, line 1:
```suggestion
# 5. Registration of a bundle-like module 
```

matks:
> I suggest to close this PR as it seems it seems to be stale now.
> 
> If we really want this topic to reach completion, I think it will be easier to create a new PR.

Closing

PR #9, opened 2020-03-23
0008 - Add Ajax error handling ADR

atomiix:


0006-ajax-error-handling.md
```diff
@@ -0,0 +1,22 @@
...
+We need to define a way to handle Ajax calls errors.
+
+As of today, Ajax calls from javascript are not handled always in the same way, sometimes even not handled at all.
+When handled, [growl](http://ksylvest.github.io/jquery-growl/) is often used this way:
+
+```javascript
+if (response.responseJSON && response.responseJSON.message) {
+    $.growl.error({message: response.responseJSON.message});
+}
+```
```
NeOMakinG: What we cold do is adding an i18n message on errors not returning JSON, if this is possible

0006-ajax-error-handling.md
```diff
@@ -0,0 +1,22 @@
...
+
+Date: 2020-03-23
+
+## Status
+
+In discussion
+
+## Context
+
+We need to define a way to handle Ajax calls errors.
```
PierreRambaud: We don't need to define a way to handle ajax calls errors. We need to define ajax calls errors :sweat_smile: 
Since we use jQuery: have a look at https://api.jquery.com/ajaxerror/
With this you can handle all ajax errors by a generic check.

```
if response has message:
  message = json message;
else:
  message = generic message;
fi

if growl:
  use growl(message)
else:
  console.log?
fi
```

NeOMakinG: That should be done on every calls (or at least on every "main" actions to avoid spamming since sometimes an actions has more than 3 calls)

matks:
It looks like this ADR is kindof stale. We need to simply choose one solution and start using it everywhere.

I suggest @eternoendless chooses the solution to use and then we can complete merge this ADR.

matks:
@atomiix  if you are OK, maybe we should close this? we were not able to manage this topic until a vote

PR #12, opened 2020-08-20
0009 - Expose js core modules to the window object to provide an interface to every developers

NeOMakinG:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  | We need to decide about a proper workflow of exposing core js modules to every developers using a global object such as window
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no

## Vote

Maintainer | Vote
---------- | ----
@eternoendless | 
@PierreRambaud | ✅ 
@matks| ✅ 
@jolelievre | ✅ 
@Progi1984 | 
@matthieu-rolland| ✅ 
@atomiix  | 
@NeOMakinG | ✅ 
@sowbiba |

### DIFF
--- a/0005-expose-js-modules-using-window-variable.md
+++ b/0005-expose-js-modules-using-window-variable.md
@@ -23,7 +23,7 @@ The idea would be to give them the possibility to initialize TranslatableInput u
 
 This means that if they are already initiated, they need to avoid a new execution.
 
-Currently, the `prestashop` object is not used in the BO, but as we wants to stick to the workflow used on FO, we would like to introduce it in the backoffice.
+Currently, the `prestashop` object is not used in the BO, but as we want to stick to the workflow used on FO, we would like to introduce it in the backoffice.
 
 @JevgenijVisockij submited a way of initializing global objects :
 
@@ -45,17 +45,33 @@ const initComponents = () => {
 }
 ```
 
+and used like :
+
+```
+$(() => {
+    window.prestashop.components.translatableInput.init();
+});
+```
+
+and maybe you could get the possibility to access to the component directly like :
+
+```
+$(() => {
+    new window.prestashop.components.translatableInput.component({localeItemSelector: '.my-selector'});
+});
+```
+
 based on a little POC, which was mainly a proof of concept : [20313](https://github.com/PrestaShop/PrestaShop/pull/20313)
 
-This means that we'll have to refacto a lot of javascripts in order to port every components into globals, or atleast begin by mosts used ones.
+This means that we'll have to refacto a lot of javascripts in order to port every components into globals, or at least begin by mosts used ones.
 
 ## Consequences
 
-What become easier :
+What becomes easier :
 
-- As a module developer, you'll be able to use every global javascript features of the core (if they are exposed inside the window objet)
+- As a module developer, you'll be able to use every global javascript features of the core (if they are exposed inside the window object)
 - As a module developer, you're no longer forced to use hard paths breaking your dev environment or CI/CD workflow
-- As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window objet, you'll be able to use them directly
+- As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window object, you'll be able to use them directly
 - We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they wants without caring of others modules. It means that we'll maybe avoid some side effects on some features.
 - Modules will be able to override core components without breaking the shop update workflow
 
--- a/0005-expose-js-modules-using-window-variable.md
+++ b/0005-expose-js-modules-using-window-variable.md
@@ -72,7 +72,7 @@ What becomes easier :
 - As a module developer, you'll be able to use every global javascript features of the core (if they are exposed inside the window object)
 - As a module developer, you're no longer forced to use hard paths breaking your dev environment or CI/CD workflow
 - As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window object, you'll be able to use them directly
-- We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they wants without caring of others modules. It means that we'll maybe avoid some side effects on some features.
+- We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they wants without caring of other modules. It means that we'll maybe avoid some side effects on some features.
 - Modules will be able to override core components without breaking the shop update workflow
 
 What could become harder :
--- a/0005-expose-js-modules-using-window-variable.md
+++ b/0005-expose-js-modules-using-window-variable.md
@@ -1,0 +1,64 @@
+# 5. Expose js modules using window variable
+
+Date: 2020-08-20
+
+## Status
+
+In discussion
+
+## Context
+
+The issue motivating this decision, and any context that influences or constrains the decision.
+Modules are unable to use translatable type without using hardlinks such as :
+
+```
+import TranslatableInput from '../../../../../admin-dev/themes/new-theme/js/components/translatable-input';
+```
+
+This path is making CI/CD harder, and also break on some development environment such as symlinks or containers.
+
+## Decision
+
+The idea would be to give them the possibility to initialize TranslatableInput using a global objet such as `window.prestashop.components.TranslatableInput.init()`.
+
+This means that if they are already initiated, they need to avoid a new execution.
+
+Currently, the `prestashop` object is not used in the BO, but as we wants to stick to the workflow used on FO, we would like to introduce it in the backoffice.
+
+@JevgenijVisockij submited a way of initializing global objects :
+
+```
+const initComponents = () => {
+  let translatableInput = null;
+  window.prestashop = {
+    components: {
+      translatableInput: {
+        init() {
+          if (translatableInput === null) {
+            translatableInput = new TranslatableInput();
+          }
+          return translatableInput;
+        },
+      },
+    }
+  }
+}
+```
+
+based on a little POC, which was mainly a proof of concept : [20313](https://github.com/PrestaShop/PrestaShop/pull/20313)
+
+This means that we'll have to refacto a lot of javascripts in order to port every components into globals, or atleast begin by mosts used ones.
+
+## Consequences
+
+What become easier :
+
+- As a module developer, you'll be able to use every global javascript features of the core (if they are exposed inside the window objet)
+- As a module developer, you're no longer forced to use hard paths breaking your dev environment or CI/CD workflow
+- As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window objet, you'll be able to use them directly
+- We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they wants without caring of others modules. It means that we'll maybe avoid some side effects on some features.
+- Modules will be able to override core components without breaking the shop update workflow
+
+What could become harder :
+
+- It will require more knowledge about the project. Before, you needed to know that TranslatableInput exist, while you'll need to also now that they are global and don't require to be imported at all.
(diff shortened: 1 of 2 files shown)


0005-expose-js-modules-using-window-variable.md
```diff
@@ -0,0 +1,64 @@
...
+
+This means that we'll have to refacto a lot of javascripts in order to port every components into globals, or atleast begin by mosts used ones.
+
+## Consequences
+
+What become easier :
+
+- As a module developer, you'll be able to use every global javascript features of the core (if they are exposed inside the window objet)
+- As a module developer, you're no longer forced to use hard paths breaking your dev environment or CI/CD workflow
+- As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window objet, you'll be able to use them directly
```
jolelievre, line 58:
```suggestion
- As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window object, you'll be able to use them directly
```

0005-expose-js-modules-using-window-variable.md
```diff
@@ -0,0 +1,64 @@
...
+- We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they wants without caring of others modules. It means that we'll maybe avoid some side effects on some features.
```
jolelievre, line 59:
```suggestion
- We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they want without caring of other modules. It means that we'll maybe avoid some side effects on some features.
```

0005-expose-js-modules-using-window-variable.md
```diff
@@ -0,0 +1,64 @@
...
+- Modules will be able to override core components without breaking the shop update workflow
+
+What could become harder :
+
+- It will require more knowledge about the project. Before, you needed to know that TranslatableInput exist, while you'll need to also now that they are global and don't require to be imported at all.
```
jolelievre, line 64: It's not that much harder, as long as we document it and refacto the core code which will provide some good examples

0005-expose-js-modules-using-window-variable.md
```diff
@@ -0,0 +1,64 @@
...
+          if (translatableInput === null) {
+            translatableInput = new TranslatableInput();
+          }
+          return translatableInput;
+        },
+      },
+    }
+  }
+}
+```
```
jolelievre, line 46: This explains how it is done, maybe we should an example of how it is used:

```js
$(() => {
    window.prestashop.components.translatableInput.init();
});
```

Also we should wait for the PR to be over, because we discussed this morning about making the component accessible as well, separately from the init method in case you want to use it differently from the default behaviour (like use other selectors in the constructor).

The idea is that you could do an alternative init via:

```js
$(() => {
    new window.prestashop.components.translatableInput.component({localeItemSelector: '.my-selector'});
});
```

The PR is not final, so we can either discuss it in this ADR, validate it and update the PR. Or the opposite, I don't care ^^

0005-expose-js-modules-using-window-variable.md
NeOMakinG: Yes, but it's harder anyway, this is one more thing to know than before, but this is the only thing I found

PierreRambaud:
> Here's what we discussed:
> 
> All PrestaShop components are bundled together and its code is available in all pages. Each controller decides which components it chooses to initialize.
> 
> There's not one, but two global variables: `window.PS` and `window.prestashop`. The first one is a namespace, the second one contains objects.
> 
> The `PS` namespace will contain classes like this `PS.Component.SomeComponent`. If you want to get a new instance of SomeComponent, you call `new PS.Component.SomeComponent(...params)`
> 
> The `prestashop` object contains two things: `initComponents()` and `components`.
> 
> `prestashop.initComponents()` is a factory method that receives an array of component names like this: `['PS.Component.SomeComponent', 'PrestaShop.Component.SomeOtherComponent']`. It will create instances of the received classnames using default parameters and will store those instances in `prestashop.components`. So that when you create a `PS.Component.SomeComponent` then you will have an instance of it available in `prestashop.components.someComponent`.
> 
> Since you have access to both constructors and components, developers are free to choose how to initialize and control their components.
> 
> If you don't want to initialize a given component with default parameters, you can always call `new PS.Component.SomeComponent(...myOwnParameters)`. You can even add it to the list of global components: `prestashop.components.someComponent = new PS.Component.SomeComponent(...)`.
> 
> If you need to apply some mutation to an already initialized component, you just get the global instance: `prestashop.components.someComponent.doSomething(...)`.

We should use only one global variable, don't forget when you put data in the `window` object, they are registered as global, this means: `window.PS` = `PS`, so having two globals variables is a risk, a module can delete one of them :/

Why not
```js
window.prestashop.Component.SomeComponent;
window.prestashop.components;
```

eternoendless:
I proposed using two variables to make it clear that one is a namespace and the other a collection.

> so having two globals variables is a risk, a module can delete one of them

I can't see it would be less risky to have a single variable having into account that modules can do whatever they like :)

PierreRambaud:
> I proposed using two variables to make it clear that one is a namespace and the other a collection.
> 
> > so having two globals variables is a risk, a module can delete one of them
> 
> I can't see it would be less risky to have a single variable having into account that modules can do whatever they like :)

The more globals variables you have, the more you have to lose :trollface: 
it can also be confusing, `PS` looks outdated compared to `prestashop`, people can think it's a deprecated variable or regarding the ecosystem, think it's an issue :sweat_smile:

eternoendless:
I wanted to make a clear difference between the two:

```
prestashop.Component.SomeComponent   <-- the component class
prestashop.components.someComponent  <-- the component instance
```

The difference is too subtle there, I'm sure people will mix them up all the time.

How about this:

```
prestashop.component.SomeComponent    <-- the class
prestashop.allComponents.someComponent <-- the instance

// alternatives
- prestashop.global.someComponent
- prestashop.instance.someComponent

**Comment by NeOMakinG** [comment]
I'm in favor of 

```
- prestashop.component.SomeComponent
- prestashop.instance.someComponent
```

It's way clearer tbh

**Comment by jolelievre** [comment]
I also agree with @NeOMakinG naming proposition. And I would add one last modif related to this naming and the `initComponents` function, we could simplify it so that it is used this way:

```js
initComponents(['TranslatableInput']);
```

instead of using a long namespace chain, since every component classes will be stored in `window.prestashop.component` anyway it will be easy to access them without some complex parsing of a namespace

This also avoid declaring the namespaced classes as global, just in the prestashop component list is enough, which also allows modules to easily add their own component

0009-expose-js-components-using-window-variable.md
```diff
@@ -0,0 +1,61 @@
...
+
+We have decided about a system which resolves in 4 concepts:
+
+1. Reusable components in BO will be available globally through `window.prestashop` (name can still be modified in short term).
+
+All PrestaShop components will be bundled together and made available in all pages using this mean. Each controller decides which components it chooses to initialize.
+
+2. Reusable components will be available as a namespace `window.prestashop.component`.
+
+The namespace will contain classes like this `prestashop.component.SomeComponent`. If you want to get a new instance of `SomeComponent`, you call new `prestashop.component.SomeComponent(...params)`
```
jolelievre, line 30:
```suggestion
The namespace will contain classes like this `prestashop.component.SomeComponent`. If you want to get a new instance of `SomeComponent`, you call `new prestashop.component.SomeComponent(...params)`
```

0009-expose-js-components-using-window-variable.md
```diff
@@ -0,0 +1,61 @@
...
+
+3. Reusable components will be available as initialized instances through `window.prestashop.instance`. These instances are initialized with default parameters.
```
jolelievre, line 32:
```suggestion
3. Reusable components will be available as initialized instances through `window.prestashop.instance`. These instances are initialized with default parameters by the `initComponent` function.
```

Just an addition because at first it feels like the component are **always** present by default, but you still need to init them.

0009-expose-js-components-using-window-variable.md
```diff
@@ -0,0 +1,61 @@
...
+
+4. A function `initComponent` available through `prestashop.component` is responsible for building `window.prestashop.instance`.
+
+### Why a namespace and a collection of instances
+
+Since you have access to both constructors and components, developers are free to choose how to initialize and control their components.
+
+If you don't want to initialize a given component with default parameters, you can always call new `prestashop.component.SomeComponent(...myOwnParameters)`.
```
jolelievre, line 40:
```suggestion
If you don't want to initialize a given component with default parameters, you can always call `new prestashop.component.SomeComponent(...myOwnParameters)`.
```

jolelievre (commented):
Just typo and one precision to avoid misunderstanding, but otherwise I validate this and vote yes.

0009-expose-js-components-using-window-variable.md
jolelievre: Oh by the way it's `initComponents` with an s

matks:
## Opening the voting phase 🎉

PierreRambaud:
LGTM

NeOMakinG:
LGTM too

0009-expose-js-components-using-window-variable.md
```diff
@@ -0,0 +1,61 @@
...
+
+All PrestaShop components will be bundled together and made available in all pages using this mean. Each controller decides which components it chooses to initialize.
+
+2. Reusable components will be available as a namespace `window.prestashop.component`.
+
+The namespace will contain classes like this `prestashop.component.SomeComponent`. If you want to get a new instance of `SomeComponent`, you call `new prestashop.component.SomeComponent(...params)`
+
+3. Reusable components will be available as initialized instances through `window.prestashop.instance`. These instances are initialized with default parameters by the `initComponents` function.
+
+4. A function `initComponent` available through `prestashop.component` is responsible for building `window.prestashop.instance`.
```
jolelievre, line 34:
```suggestion
4. A function `initComponents` available through `prestashop.component` is responsible for building `window.prestashop.instance` and is used like this `prestashop.component.initComponents(['SomeComponent'])` (the list of component names must match a component present in `prestashop.component`).
```

jolelievre:
LGTM too:
sorry last modif though 😅
https://github.com/PrestaShop/ADR/pull/12/files#r479310822

matthieu-rolland:
LGTM

matthieu-rolland (approved):

jolelievre (approved):

matks:
As we have reached majority, this ADR is accepted.

matks:
Thank you everybody !

0009-expose-js-components-using-window-variable.md
```diff
@@ -0,0 +1,61 @@
...
+
+- As a module developer, you'll be able to use every global javascript features of the core (if they are exposed inside the window object)
+- As a module developer, you're no longer forced to use hard paths breaking your dev environment or CI/CD workflow
+- As a core developer, you're no longer forced to import every modules you need, if they are exposed in the window object, you'll be able to use them directly
+- We now provide a way to avoid multiple executions of the same feature, while before, every modules was using them as they wants without caring of other modules. It means that we'll maybe avoid some side effects on some features.
+- Modules will be able to override core components without breaking the shop update workflow
+
+What could become harder :
+
+- It will require more knowledge about the project. Before, you needed to know that TranslatableInput exist, while you'll need to also now that they are global and don't require to be imported at all.
```
sowbiba, line 61: ... also know that ...

matks: Nice catch 😉 

Will you submit a PR or I do it ?

NeOMakinG: Damn

PR #16, opened 2021-01-05
0011 - Use constants for configuration variables

JevgenijVisockij:


matks:
ADR added in table https://github.com/PrestaShop/ADR/blob/master/README.md

0011-use-constantas-for-configuration-variables.md
```diff
@@ -0,0 +1,35 @@
...
+
+Date: 2020-01-05
+
+## Status
+
+In discussion
+
+## Context
+
+Configuration names in PrestaShop are used as hardcoded strings. It makes it harder to figure out
```
matks, line 11: Could you provide a few examples of such configuration names and how they are used in the code ? (just to make it easier for reader to understand the topic, this ADR also serves as documentation for people in the future)

0011-use-constantas-for-configuration-variables.md
```diff
@@ -0,0 +1,35 @@
...
+It already exists for all purposes concerning usage of configurations. Creation of the new class just to store
+constants is not worthwhile in this case.
+
+## First implementation
+
+https://github.com/PrestaShop/PrestaShop/pull/22672
+
+## Consequences
+
+What becomes easier :
```
matks, line 28: To be honest, we must also list the drawbacks 😄 . This means for example we need to `require` or `use` the Configuration class everywhere.

JevgenijVisockij: Added 2 drawbacks, couldn't think of more at the moment.

0011-use-constantas-for-configuration-variables.md
JevgenijVisockij: Updated the PR to include examples.

matks (approved):
PR looks good to me. Complete. Ready to discuss and vote.

matks:
@PrestaShop/prestashop-core-developers Can you check this and ask questions if you have? Else we can start the voting phase.

Progi1984:
@matks You should ping @PrestaShop/prestashop-maintainers. May we should create a meeting for all ADRs?

matthieu-rolland (approved):
Being working on the configuration forms at the moment, I wholeheartedly agree with this ADR.

About the drawbacks, they are minor in my opinion, this could be implemented progressively.

0011-use-constantas-for-configuration-variables.md
```diff
@@ -0,0 +1,54 @@
...
+## Context
+
+Configuration names in PrestaShop are used as hardcoded strings.
+For example:
+    
+    PS_SSL_ENABLED
+    PS_SHOP_ENABLE
+ 
+It makes it harder to figure out which configurations are available, also leaves possibility for mistyping configuration name without noticing.
+
```
matthieu-rolland, line 18: Also, the names of configuration form fields don't always match the names of configuration names, for example, 'PS_SHOP_ENABLE` is linked to form field `enable_shop` (and not `shop_enable` as we would intuitively guess). 

Using constants would make it less error-prone.

kpodemski:
Test

kpodemski:
I like the idea, I'm not sure if main `Configuration` class is good for those statics,

`Configuration::get(Configuration:PS_SHOP_NAME);`
`Configuration::get(ConfigKey::PS_SHOP_NAME);

```
use ConfigurationKeys as Key;
`Configuration::get(Key::PS_SHOP_NAME);
```

I don't think it's that important, I'm thinking aloud :)

PierreRambaud:
> I like the idea, I'm not sure if main `Configuration` class is good for those statics,
> 
> `Configuration::get(Configuration:PS_SHOP_NAME);`
> `Configuration::get(ConfigKey::PS_SHOP_NAME);
> 
> ```
> use ConfigurationKeys as Key;
> `Configuration::get(Key::PS_SHOP_NAME);
> ```
> 
> I don't think it's that important, I'm thinking aloud :)

As I said in the PR, I'm like @kpodemski, using ConfigurationCore class isn't the best one (imho). I'm not even sure about the `ConfigurationKeys` class too :sweat_smile: but maybe in the adapter?

PR #17, opened 2021-01-13
Establish rules for form processing

JevgenijVisockij:


saulaski:
@JevgenijVisockij you make a serious point about form usage.

In my humble opinion:
1. to keep URL consistent, the same route should handle both form display (GET) and processing (POST/PUT)
2. in a controller the form should be created and processed in the same action method
3. redirection should only happen after the form is successfully processed, no matter what
4. sounds kinda weird to let the DataProvider validate the form, whereas form constraints or domain constraints (class annotations) are way more standard

MeKeyCool (changes requested):
@JevgenijVisockij thanks for your interest to Prestashop and for your efforts to make Prestashop great :)

I don't think this request is an " Architectural Decision Record" but more a UX/UI / product request that may imply some architectural optimizations (if a generic behavior is found)

I recommend to create an issue or to discuss with the product-team :)

PR #24, opened 2022-01-07
0016 -  Use variadic to type-hint PHP arrays

Amoifr:
| Questions         | Answers
| Description?      | New ADR for Variadic
| Type?             | improvement
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| How to test?      | 
| Possible impacts? |


0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
...
+?>
+
+it returns 6
+```
+
+when you accept variadic
+```
+<?php
+function f($req, $opt = null, ...$params) {
+    // $params est un tableau contenant les arguments restants.
```
NeOMakinG, line 34:
```suggestion
    // $params is an array containing the remaining arguments 
```

0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
...
+
+it returns 6
+```
+
+when you accept variadic
+```
+<?php
+function f($req, $opt = null, ...$params) {
+    // $params is an array containing the remaining arguments 
+    printf('$req: %d; $opt: %d; Nombre d\'arguments : %d'."\n",
```
NeOMakinG, line 35: Same here, you need to use english everywhere

Amoifr, line 35:
```suggestion
    printf('$req: %d; $opt: %d; Number of arguments : %d'."\n",
```

0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
...
+f(1, 2, 3, 4);
+f(1, 2, 3, 4, 5);
+?>
+
+it returns 
+$req: 1; $opt: 0; Nombre d'arguments : 0
+$req: 1; $opt: 2; Nombre d'arguments : 0
+$req: 1; $opt: 2; Nombre d'arguments : 1
+$req: 1; $opt: 2; Nombre d'arguments : 2
+$req: 1; $opt: 2; Nombre d'arguments : 3
```
Amoifr, line 51:
```suggestion
$req: 1; $opt: 0; Number of arguments : 0
$req: 1; $opt: 2; Number of arguments : 0
$req: 1; $opt: 2; Number of arguments : 1
$req: 1; $opt: 2; Number of arguments : 2
$req: 1; $opt: 2; Number of arguments : 3
```

0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
...
+      ...
+    }
+
+    $this->json($this->formatProductsForAssociation($products));
+
+```
+
+It allows to type without creating a DTO for a simple collection !
+
+"Magie du Pestacle !!!"
```
Amoifr, line 83:
```suggestion
"Magie du Pestacle !!!" => "show magic"
```

NeOMakinG:
Even if it's PHP and I'm a front-end developer, I have to agree with this, no kind of types like in TS with `Array<Products>` in PHP?

Amoifr:
> 



> @NeOMakinG The best way is to pass a DTO as a parameter, but this is a workaround if you just have one dimension.

0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Status
+
+Pending
+
+## Context
+
+Using methods with array in signature force to type the parameters in phpDocs. It exists a workaround to force the type of the items in the array, and have astronger typing :
+
+THE VARIADIC
```
matks, line 13:
```suggestion
THE [VARIADIC](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list)
```

0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
...
+      ...
+    }
+
+    $this->json($this->formatProductsForAssociation($products));
+
+```
+
+It allows to type without creating a DTO for a simple collection !
+
+"Magie du Pestacle !!!" => "show magic"
```
matks, line 83: Maybe we dont want this part to remain here for posterity 😄

matks (approved):
Some improvements needed about the text and details but I approve the idea

0016 - Variadic usage.md
```diff
@@ -0,0 +1,95 @@
+# 15. Variadic usage in method params
```
matks, line 1:
```suggestion
# 16.  Use variadic to type-hint PHP arrays
```

matks:
ADR added to README table https://github.com/PrestaShop/ADR/blob/master/README.md

0016 - Variadic usage.md
Amoifr: good for me

0016 - Variadic usage.md
Amoifr: I've tried, ... and I've been catched

0016 - Variadic usage.md
```diff
@@ -0,0 +1,93 @@
...
+    $this->json($this->formatProductsForAssociation($products));
+
+```
+
+It allows to type without creating a DTO for a simple collection !
+
+
+## Decision
+
+
```
jolelievre, line 86: I agree with using this kind of implementation However, we should define when it should be used and when we should use an array instead For example one of the big drawbacks of variadic is that it needs to be the last parameter (which makes sense since there is no way of knowing how many parameters will be used)

So it's not ideal to combine it with optional parameters since you need to explicitly set the optional parameters before providing your variadic which is not good for usage and/or readability

jolelievre: For that we need to define a set of rules for its usage to avoid creating useless discussions during PR review

jolelievre: One other limitation you told me I believe is regarding using two variadic with two different types:

`private function formatProductsForAssociation(ProductForAssociation ...$productsForAssociation, CombinationForAssociation ...$combinationsForAssociation): array`

If I remember correctly this can be used in PHP8 but not in previous versions so we can't accept this kind of use cases

jolelievre: A suggestion of rules:

```suggestion
- we don't use variadic if the method has optional parameters but the variadic is not
- we don't use multiple variadics with different types because it's not compatible with PHP<8
```

jolelievre: I didn't think everything through but that's a beginning

0016 - Variadic usage.md
```diff
@@ -0,0 +1,93 @@
...
+it returns 
+$req: 1; $opt: 0; Number of arguments : 0
+$req: 1; $opt: 2; Number of arguments : 0
+$req: 1; $opt: 2; Number of arguments : 1
+$req: 1; $opt: 2; Number of arguments : 2
+$req: 1; $opt: 2; Number of arguments : 3
+```
+
+In our world, mixing the two ways
```
jolelievre, line 54: Just a suggestion about redaction, it might be easier to understand if you first display the old way, and then the new way Starting the explanation with the new syntax first, which they may not know, is confusing as you don't have the context If you show the old context first at least people are in their comfort zone ready to accept the new one

0016 - Variadic usage.md
```diff
@@ -0,0 +1,93 @@
...
+
+```
+    private function formatProductsForAssociation(ProductForAssociation ...$productsForAssociation): array
```
jolelievre, line 57: You only described this as method parameters but I believe it's usable for return as well no? The usage of this new feature will only have its real value if we use them both ways: input and output

jolelievre (commented):
I'm in favor of this, although I have a few comments

sowbiba (commented):
It's kind of confusing since, according to your description, this operator can have 2 different meaning depending if it's typed or not, IF I understand well

function f($a, $b, ...$c)
means that you can pass any number of arguments to the function, when it's more than 2, they will be in the variable $c

f(1, 2, $d, $e, $f, $g) => count($c) in the function will give 4



But with function f($a, $b, Product ...$c) 

you cannot do f(1, 2, $d, $e, $f, $g)  even if $d, $e, $f, $g are Product
but you can do f(1, 2, [$d, $e, $f, $g])

I will look further to PHP documentation

sowbiba:
> It's kind of confusing since, according to your description, this operator can have 2 different meaning depending if it's typed or not, IF I understand well
> 
> function f($a, $b, ...$c) means that you can pass any number of arguments to the function, when it's more than 2, they will be in the variable $c
> 
> f(1, 2, $d, $e, $f, $g) => count($c) in the function will give 4
> 
> But with function f($a, $b, Product ...$c)
> 
> you cannot do f(1, 2, $d, $e, $f, $g) even if $d, $e, $f, $g are Product but you can do f(1, 2, [$d, $e, $f, $g])
> 
> I will look further to PHP documentation

Or maybe in your 2nd example f(Product ...$products) is not equivalent to f(array $products)

saulaski:
> Even if it's PHP and I'm a front-end developer, I have to agree with this, no kind of types like in TS with `Array<Products>` in PHP?

Not yet unfortunately :/

0016 - Variadic usage.md
saulaski: Unfortunately variadic arguments are only used as input. But as output we could use a collection DTO that implements `\iterable` or `\IteratorAggregate` interface.

saulaski:
In case BC break is an issue, I suggest a simple trick:
```
public function setItems(array $items): self
{
    $this->items = [];

    return $this->addItem(...$items);
}

protected function addItem(Item ...$items): self
{
    foreach ($items as $item) {
        $this->items[] = $item;
    }

    return $this;
}
```

saulaski:
> > It's kind of confusing since, according to your description, this operator can have 2 different meaning depending if it's typed or not, IF I understand well
> > function f($a, $b, ...$c) means that you can pass any number of arguments to the function, when it's more than 2, they will be in the variable $c
> > f(1, 2, $d, $e, $f, $g) => count($c) in the function will give 4
> > But with function f($a, $b, Product ...$c)
> > you cannot do f(1, 2, $d, $e, $f, $g) even if $d, $e, $f, $g are Product but you can do f(1, 2, [$d, $e, $f, $g])
> > I will look further to PHP documentation
> 
> Or maybe in your 2nd example f(Product ...$products) is not equivalent to f(array $products) ?

Variadic arguments are always received as an array in the function body. This array contains as many items as variadic arguments provided.
If you need to pass items in an array as variadic arguments, just use the spread operator : `f(1, 2, ...$products)`

matks:
Let's try to push this topic this week so what we can vote next week \o/

PR #25, opened 2022-03-24
0017 - Add our backward compatibility promise

PierreRambaud:
| Questions         | Answers
| Description?      | Try to explain our backward compatibility promise

Votes

## Vote

Maintainer | Vote (✅  or ❌)
---------- | ----
@atomiix  | 
@eternoendless | ✅
@jolelievre | ✅
@kpodemski | ✅
@matks |  ✅ 
@matthieu-rolland | ✅
@NeOMakinG | ✅
@PululuK | 
@sowbiba |

### DIFF
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -10,7 +10,7 @@ Accepted
 
 PrestaShop is a development platform used by developers and integrators to create e-commerce sites. Developers can create other software (including modules & themes) to interact with PrestaShop, extend it, or modify its behavior. For that, they rely on software interfaces which are commonly referred to as "APIs". Whenever any of these interfaces change, existing third-party software must be adapted to work with the new interface -- this is called a "backward-incompatible change", "breaking change" ("BC") or "backward compatibility break" ("BC break"). 
 
-Developing a modular and extensible platform like PrestaShop requires a delicate balance between allowing the software to evolve in the optimal way, and keeping its interafaces as they are, for as long as possible, in order to reduce impact on third-party software. For that, it is important to clarify what belongs to the API and must be protected from changes, and what part of the software is private and can be modified freely.
+Developing a modular and extensible platform like PrestaShop requires a delicate balance between allowing the software to evolve in the optimal way, and keeping its interfaces as they are, for as long as possible, in order to reduce impact on third-party software. For that, it is important to clarify what belongs to the API and must be protected from changes, and what part of the software is private and can be modified freely.
 
 Further reading: [The Pain Points](https://build.prestashop.com/news/prestashop-in-2019-and-beyond-part-2-pain-points/#no-clearly-defined-api)
 
@@ -18,9 +18,9 @@ Further reading: [The Pain Points](https://build.prestashop.com/news/prestashop-
 
 PrestaShop is committed to follow the [SemVer convention](https://semver.org/), which states that backward incompatible changes are only allowed in major versions.
 
-Only exceptional circumstances (e.g. fixing a security issue) can allow introducing backward incompatible changes in minor or patch versions. Whenever this happens, the maintainer team will make sure to provide a clear list of the affected components.
+Only exceptional circumstances (e.g. fixing a security issue, critical bug, ...) can allow introducing backward-incompatible changes in minor or patch versions. Whenever this happens, the maintainer team will make sure to provide a clear list of the affected components.
 
-The following lines detail or backward compatibility promise, and describe what changes are allowed or not in minor or patch versions. Further restrictions (or allowances) might be added in the future. 
+The following lines detail our backward compatibility promise, and describe what changes are allowed or not in minor or patch versions. Further restrictions (or allowances) might be added in the future. 
 
 ### Packages
 
@@ -37,7 +37,7 @@ Our backward compatibility promise only applies to production packages, except f
 
 Note: "Symfony" is mentioned below as relating to the Symfony framework and its components, including the service container.
 
-| Type of Change                                                      | Allowed |
+| Type of Change                                                      | Allowed        |
 |---------------------------------------------------------------------|----------------|
 | Remove an interface/class                                           | No             |
 | Remove public or protected method [^1]                              | No             |
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -4,7 +4,7 @@ Date: 2022-03-24
 
 ## Status
 
-N/A
+Accepted
 
 ## Context
 
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -30,7 +30,7 @@ Backward compatibility guarantees only apply to the production packages except f
 | Interface/class removal                                        | No             |
 | Public and protected method removal [1]                        | No             |
 | Change/remove service name                                     | No             |
-| Composer dependencies without BC breaks                        | Yes            |
+| Composer dependencies upgrade without BC breaks                        | Yes            |
 | Composer dependencies used a lot                               | No             |
 | Composer dev dependencies                                      | Yes            |
 | Change Hook name                                               | No             |
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -70,7 +70,7 @@ PrestaShop uses Commands and Queries. Specific rules apply to these PHP objects 
 | Type of Change                                     | Change Allowed |
 |----------------------------------------------------|----------------|
 | Rename                                             | No             |
-| Default parameter values                           | No             |
+| Modify default parameter values                           | No             |
 | Return type                                        | No             |
 | Add parameter                                      | No             |
 | Add optional parameter at the end of the args list | Yes            |
@@ -161,7 +161,7 @@ All directories name and structure are part of the public API until the split be
 | Change field default value                  | No             |
 | Change/remove table name                    | No             |
 | Add new optional field to a table           | Yes            |
-| Introducting a required field               | No             |
+| Add a required field               | No             |
 | Change/remove constraints                   | Yes            |
 |                                             |                |
 
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -15,7 +15,7 @@ This page describes rules and best practices for backward compatible development
 
 The PrestaShop code base is composed of two different types of packages:
 
- - **Production packages:** these are packages that are shipped 
+ - **Production packages:** these are packages that are shipped in the final release build
  - **Development packages:** these are tools that can be used by third-party developers to lint, test, format or build whatever they want. They are used as npm or composer dependencies.
 
 Backward compatibility guarantees only apply to the production packages except for modules which are not part of the public API.
@@ -29,10 +29,11 @@ Backward compatibility guarantees only apply to the production packages except f
 |----------------------------------------------------------------|----------------|
 | Interface/class removal                                        | No             |
 | Public and protected method removal [1]                        | No             |
-| Change/remove service name                                     | No             |
+| Change service name                                     | No             |
+| Remove service                                          | No             |
 | Composer dependencies upgrade without BC breaks                        | Yes            |
 | Composer dependencies used a lot                               | No             |
-| Composer dev dependencies                                      | Yes            |
+| Updating/removing composer dev dependencies                                      | Yes            |
 | Change Hook name                                               | No             |
 | Remove hook parameters                                         | No             |
 | Add hook parameters                                            | Yes            |
@@ -48,7 +49,7 @@ Backward compatibility guarantees only apply to the production packages except f
 
 | Type of Change            | Change Allowed |
 |---------------------------|----------------|
-| Queries                   | No             |
+| Change/remove Query constructor parameters                   | No             |
 | QueryResult (constructor) | yes            |
 | QueryResult (getters)     | No             |
 |                           |                |
@@ -61,7 +62,7 @@ Backward compatibility guarantees only apply to the production packages except f
 | Default parameter values | No             |
 | Return type              | No             |
 | Add parameter            | No             |
-| Force strict typing      | No             |
+| Add strict typing      | No             |
 |                          |                |
 
 #### Symfony Controller actions
@@ -74,7 +75,7 @@ Experimental Features and code should be marked with the `@internal` tags.
 
 #### Security issues
 
-Note that backward compatibility breaks are tolerated if they are required to fix a security issue.
+Backward compatibility breaks are tolerated if they are required to fix a security issue.
 
 ### Front
 
@@ -86,9 +87,10 @@ Note that backward compatibility breaks are tolerated if they are required to fi
 | `id` and `class`                       | No             |
 | Npm dev dependencies                   | Yes            |
 | Npm dependencies                       | No             |
-| Functions not exposed to the DOM       | Yes            |
-| Functions directly exposed to the DOM  | No             |
-| Change/remove template file name       | No             |
+| Functions not exposed to the global scope       | Yes            |
+| Functions directly exposed to the global scope  | No             |
+| Remove a template file                 | No             |
+| Change template path                   | No             |
 | Change/remove javascript file name [2] | Yes            |
 |                                        |                |
 
@@ -143,7 +145,7 @@ All directories name and structure are part of the public API until the split be
 | Change field default value                  | No             |
 | Change/remove table name                    | No             |
 | Add new optional field to a table           | Yes            |
-| Introducting a required filed               | No             |
+| Introducting a required field               | No             |
 | Change/remove constraints                   | Yes            |
 |                                             |                |
 
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -8,79 +8,82 @@ N/A
 
 ## Context
 
-The following lines describe what changes are considered backward compatible and what changes are not.
+PrestaShop is a development platform used by developers and integrators to create e-commerce sites. Developers can create other software (including modules & themes) to interact with PrestaShop, extend it, or modify its behavior. For that, they rely on software interfaces which are commonly referred to as "APIs". Whenever any of these interfaces change, existing third-party software must be adapted to work with the new interface -- this is called a "backward-incompatible change", "breaking change" ("BC") or "backward compatibility break" ("BC break"). 
 
-PrestaShop attempts to follow [SemVer](https://semver.org/) which states that not-backward compatible changes are only allowed in major versions.
+Developing a modular and extensible platform like PrestaShop requires a delicate balance between allowing the software to evolve in the optimal way, and keeping its interafaces as they are, for as long as possible, in order to reduce impact on third-party software. For that, it is important to clarify what belongs to the API and must be protected from changes, and what part of the software is private and can be modified freely.
 
-Some circumstances might require to break this promise. Whenever such a usecase happens, the maintainer team will make sure to provide a clear list of the not backward compatible changes introduced in a patch or minor version.
+Further reading: [The Pain Points](https://build.prestashop.com/news/prestashop-in-2019-and-beyond-part-2-pain-points/#no-clearly-defined-api)
 
-For example backward compatibility breaks are allowed if they are required to fix a security issue.
+## Decision
+
+PrestaShop is committed to follow the [SemVer convention](https://semver.org/), which states that backward incompatible changes are only allowed in major versions.
+
+Only exceptional circumstances (e.g. fixing a security issue) can allow introducing backward incompatible changes in minor or patch versions. Whenever this happens, the maintainer team will make sure to provide a clear list of the affected components.
+
+The following lines detail or backward compatibility promise, and describe what changes are allowed or not in minor or patch versions. Further restrictions (or allowances) might be added in the future. 
 
 ### Packages
 
 The PrestaShop code base is composed of two different types of packages:
 
- - **Production packages:** these are packages that are shipped in the final release build
- - **Development packages:** these are tools that can be used by third-party developers to lint, test, format or build whatever they want. They are used as npm or composer dependencies.
+ - **Production packages:** packages that are shipped in the final release build
+ - **Development packages:** tools that can be used by third-party developers to lint, test, format or build whatever they want. They are used as npm or composer dependencies.
 
-Backward compatibility guarantees only apply to the production packages except for modules which are not part of the public API.
+Our backward compatibility promise only applies to production packages, except for modules which are not part of the public API.
 
 ### PHP
 
 #### General purpose
 
-Note: Symfony is mentioned below as rely on Symfony framework and its container
+Note: "Symfony" is mentioned below as relating to the Symfony framework and its components, including the service container.
 
-| Type of Change                                                      | Change Allowed |
+| Type of Change                                                      | Allowed |
 |---------------------------------------------------------------------|----------------|
-| Interface/class removal                                             | No             |
-| Public and protected method removal [1]                             | No             |
-| Change Symfony service name                                         | No             |
-| Remove Symfony service                                              | No             |
-| Change Hook name                                                    | No             |
+| Remove an interface/class                                           | No             |
+| Remove public or protected method [^1]                              | No             |
+| Change a Symfony service name                                       | No             |
+| Remove a Symfony service                                            | No             |
+| Change a Hook name                                                  | No             |
 | Remove hook parameters                                              | No             |
 | Add hook parameters                                                 | Yes            |
 | Change value of a constant                                          | Yes            |
 | Remove or rename constants                                          | No             |
 | Adding argument to an Symfony event                                 | Yes            |
-| Modifying the type of thrown exceptions                             | No             |
+| Modify the type of thrown exceptions (except to a parent class)     | No             |
 | Throw a new type of exception from an existing method               | Yes            |
-| Adding a optional constructor parameter at the end of the args list | Yes            |
-|                                                                     |                |
+| Add an optional constructor parameter at the end of the args list   | Yes            |
 
 #### Composer dependencies
 
-| Type of Change                                                              | Change Allowed |
-|-----------------------------------------------------------------------------|----------------|
-| Upgrade composer dependencies that do not introduce BC breaks               | Yes            |
-| Composer dependencies with a strong impact on userland code and modules [2] | No             |
-| Updating/removing composer dev dependencies                                 | Yes            |
+| Type of Change                                                                     | Allowed |
+|------------------------------------------------------------------------------------|----------------|
+| Add a production package                                                           | Yes            |
+| Update a production package to a version that does not introduce BC breaks         | Yes            |
+| Update a production package with a strong impact on userland code and modules [^2] | No             |
+| Update/remove a dev package                                                        | Yes            |
 
 #### CQRS
 
-PrestaShop uses Commands and Queries. Specific rules apply to these PHP objects that define a contract.
+PrestaShop uses Commands and Queries which are part of the public API. Specific rules apply to these objects.
 
-| Type of Change                              | Change Allowed |
+| Type of Change                              | Allowed |
 |---------------------------------------------|----------------|
 | Change a QueryResult constructor            | Yes            |
-|                                             |                |
 
 #### PHP methods and functions
 
-| Type of Change                                     | Change Allowed |
+| Type of Change                                     | Allowed |
 |----------------------------------------------------|----------------|
 | Rename                                             | No             |
-| Modify default parameter values                           | No             |
-| Return type                                        | No             |
+| Modify default parameter values                    | No             |
+| Change return type                                 | No             |
 | Add parameter                                      | No             |
 | Add optional parameter at the end of the args list | Yes            |
 | Add strict typing                                  | No             |
-|                                                    |                |
 
+#### Symfony service constructors
 
-#### Symfony services constructors
-
-As classes whose instantiation is delegated to the Symfony container are not supposed to be instantiated otherwise, changing the constructor of those classes is allowed.
+Since classes whose instantiation is delegated to the Symfony service container are not supposed to be instantiated individually, changing the constructor of those classes is allowed.
 
 #### Symfony Controller actions
 
@@ -96,52 +99,53 @@ Experimental code is not part of the public API.
 
 #### Back office theme
 
-| Type of Change                                            | Change Allowed |
+| Type of Change                                            | Allowed |
 |-----------------------------------------------------------|----------------|
 | HTML markup                                               | Yes            |
-| `id` and `class`                                          | No             |
-| Npm dev dependencies                                      | Yes            |
+| Modify `id` or `class`                                    | No             |
+| Add `id` or `class`                                       | Yes            |
+| Add/remove/update npm dev dependencies                    | Yes            |
 | Upgrade npm dependencies that do not introduce BC breaks  | Yes            |
 | Functions not exposed to the global scope                 | Yes            |
 | Functions directly exposed to the global scope            | No             |
 | Remove a template file                                    | No             |
 | Change template path                                      | No             |
-| Change/remove javascript file name [3]                    | Yes            |
-|                                                           |                |
+| Change/remove javascript file name [^3]                   | Yes            |
+
 
 #### Front office
 
-| Type of Change                                            | Change Allowed |
+| Type of Change                                            | Allowed |
 |-----------------------------------------------------------|----------------|
 | Change/remove routes                                      | No             |
 | Add routes                                                | Yes            |
-| Javascript Core functions (name, parameters, return type) | No             |
-| Remove injected variables into template engines           | No             |
+| Change Javascript Core functions (name, parameters, return type) | No             |
+| Remove variables injected into template engines           | No             |
 | Add variables into template engines                       | Yes            |
 | Change/remove template file name                          | No             |
-| Change/remove javascript file name [3]                    | Yes            |
-|                                                           |                |
+| Change/remove javascript file name [^3]                   | Yes            |
+
 
 #### Theme
 
-| Type of Change                                            | Change Allowed |
+| Type of Change                                            | Allowed |
 |-----------------------------------------------------------|----------------|
 | HTML markup                                               | Yes            |
-| `id` and `class`                                          | No             |
-| Npm dev dependencies                                      | Yes            |
-| UUpgrade npm dependencies that do not introduce BC breaks | Yes            |
-| Functions not exposed to the DOM                          | Yes            |
-| Functions directly exposed to the DOM                     | No             |
+| Add `id` or `class`                                       | Yes            |
+| Add/remove/update npm dev dependencies                    | Yes            |
+| Upgrade npm dependencies that do not introduce BC breaks  | Yes            |
+| Change/remove functions not exposed to the DOM            | Yes            |
+| Change/remove functions directly exposed to the DOM       | No             |
 | Change/remove template file name                          | No             |
-|                                                           |                |
+
 
 ### Directory structure
 
-All directories name and structure are part of the public API until the split between public and private directories is done.
+All directory names and structure are part of the public API. In the future, a split between public and private directories will be determined.
 
 ### Webservices
 
-| Type of Change                     | Change Allowed |
+| Type of Change                     | Allowed |
 |------------------------------------|----------------|
 | Change/remove routes               | No             |
 | Add routes                         | Yes            |
@@ -150,22 +154,20 @@ All directories name and structure are part of the public API until the split be
 | Remove parameters                  | No             |
 | Add optional parameters            | Yes            |
 | Add required parameters            | No             |
-|                                    |                |
 
 
 ### Database structure
 
-| Type of Change                              | Change Allowed |
+| Type of Change                              | Allowed |
 |---------------------------------------------|----------------|
 | Change/remove field type or name to a table | No             |
 | Change field default value                  | No             |
 | Change/remove table name                    | No             |
 | Add new optional field to a table           | Yes            |
-| Add a required field               | No             |
+| Add a required field                        | No             |
 | Change/remove constraints                   | Yes            |
-|                                             |                |
 
 
-[1] Not if the class is tagged as `@internal` or `@experimental`
-[2] Examples: Doctrine, Guzzle...
-[3] Only if it's not a production file
+[^1]: Not if the class is tagged as `@internal` or `@experimental`
+[^2]: Examples: Doctrine, Guzzle...
+[^3]: Only if it's not a production file
--- a/0017-backward-compatibility-promise.md
+++ b/0017-backward-compatibility-promise.md
@@ -1,0 +1,152 @@
+# 17. Our Backward Compatibility Promise
+
+Date: 2022-03-24
+
+## Status
+
+N/A
+
+## Context
+
+If you are a theme or module developer, the following guidelines will help you to ensure smooth upgrades to all futur minor releases.
+This page describes rules and best practices for backward compatible development.
+
+### Packages
+
+The PrestaShop code base is composed of two different types of packages:
+
+ - **Production packages:** these are packages that are shipped 
+ - **Development packages:** these are tools that can be used by third-party developers to lint, test, format or build whatever they want. They are used as npm or composer dependencies.
+
+Backward compatibility guarantees only apply to the production packages except for modules which are not part of the public API.
+
+
+### PHP
+
+#### General purpose
+
+| Type of Change                                                 | Change Allowed |
+|----------------------------------------------------------------|----------------|
+| Interface/class removal                                        | No             |
+| Public and protected method removal [1]                        | No             |
+| Change/remove service name                                     | No             |
+| Composer dependencies without BC breaks                        | Yes            |
+| Composer dependencies used a lot                               | No             |
+| Composer dev dependencies                                      | Yes            |
+| Change Hook name                                               | No             |
+| Remove hook parameters                                         | No             |
+| Add hook parameters                                            | Yes            |
+| Change value of a constant                                     | Yes            |
+| Remove or rename constants                                     | No             |
+| Adding argument to an event                                    | Yes            |
+| Modifying the types of thrown exceptions                       | No             |
+| Throw a new type of exception from an existing method          | Yes            |
+| Adding a optional constructor parameter at the end of the list | Yes            |
+|                                                                |                |
+
+#### CQRS
+
+| Type of Change            | Change Allowed |
+|---------------------------|----------------|
+| Queries                   | No             |
+| QueryResult (constructor) | yes            |
+| QueryResult (getters)     | No             |
+|                           |                |
+
+#### Methods and functions
+
+| Type of Change           | Change Allowed |
+|--------------------------|----------------|
+| Rename                   | No             |
+| Default parameter values | No             |
+| Return type              | No             |
+| Add parameter            | No             |
+| Force strict typing      | No             |
+|                          |                |
+
+#### Symfony Controller actions
+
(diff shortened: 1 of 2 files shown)


0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+### PHP
+
+#### General purpose
+
+| Type of Change                                                 | Change Allowed |
+|----------------------------------------------------------------|----------------|
+| Interface/class removal                                        | No             |
+| Public and protected method removal [1]                        | No             |
+| Change/remove service name                                     | No             |
+| Composer dependencies without BC breaks                        | Yes            |
```
matks, line 33:
```suggestion
| Composer dependencies upgrade without BC breaks                        | Yes            |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| Composer dependencies used a lot                               | No             |
```
matks, line 34:
```suggestion
| Composer dependencies used a lot                               | No             |
```
We might need to find a better wording that "used a lot" 😄

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| Throw a new type of exception from an existing method          | Yes            |
+| Adding a optional constructor parameter at the end of the list | Yes            |
+|                                                                |                |
+
+#### CQRS
+
+| Type of Change            | Change Allowed |
+|---------------------------|----------------|
+| Queries                   | No             |
+| QueryResult (constructor) | yes            |
```
matks, line 52: Maybe we need an example so that people understand what we mean by "changing the constructor is allowed, changing getters is not".

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+
+#### Methods and functions
+
+| Type of Change           | Change Allowed |
+|--------------------------|----------------|
+| Rename                   | No             |
+| Default parameter values | No             |
+| Return type              | No             |
+| Add parameter            | No             |
+| Force strict typing      | No             |
```
matks, line 64:
```suggestion
| Enable strict typing      | No             |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+
+| Type of Change                         | Change Allowed |
+|----------------------------------------|----------------|
+| HTML markup                            | Yes            |
+| `id` and `class`                       | No             |
+| Npm dev dependencies                   | Yes            |
+| Npm dependencies                       | No             |
+| Functions not exposed to the DOM       | Yes            |
+| Functions directly exposed to the DOM  | No             |
+| Change/remove template file name       | No             |
```
matks, line 91: Maybe we should add "move file"?

matks (commented):
Some first comments :) but nice job

0017-backward-compatibility-promise.md
PierreRambaud: In fact, you can enable it without forcing it. If you enable it without the declare strict type, it's a simple cast.

matks: We need to make sure that people understand what we mean by the above. Either we find a unambiguous way to mark it or we simply provide an example 😊

0017-backward-compatibility-promise.md
PierreRambaud: yes, maybe correctly list the dependencies

0017-backward-compatibility-promise.md
PierreRambaud: I would have the advice from @jolelievre here

0017-backward-compatibility-promise.md
PierreRambaud: Why not

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+## Context
+
+If you are a theme or module developer, the following guidelines will help you to ensure smooth upgrades to all futur minor releases.
+This page describes rules and best practices for backward compatible development.
+
+### Packages
+
+The PrestaShop code base is composed of two different types of packages:
+
+ - **Production packages:** these are packages that are shipped
```
jolelievre, line 18:
```suggestion
 - **Production packages:** these are packages that are shipped in the final release build
```

0017-backward-compatibility-promise.md
jolelievre: Indeed, I'm not sure what you mean by this line?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| Composer dev dependencies                                      | Yes            |
+| Change Hook name                                               | No             |
+| Remove hook parameters                                         | No             |
+| Add hook parameters                                            | Yes            |
+| Change value of a constant                                     | Yes            |
+| Remove or rename constants                                     | No             |
+| Adding argument to an event                                    | Yes            |
+| Modifying the types of thrown exceptions                       | No             |
+| Throw a new type of exception from an existing method          | Yes            |
+| Adding a optional constructor parameter at the end of the list | Yes            |
```
jolelievre, line 44: What about non-optional parameters? I agree that we should introduce new parameters as optional as much as possible However it's not always possible I'm especially thinking about services which are built via DI the most important is that their interface remain unchanged but it they need another dependency to fix a bug does it mean we don't fix the bug to preserve backward compatibility

This is especially true for CQRS handlers (query and/or command) as they are mere proxies/middleware between the command and its execution, so we often need to add or change their dependencies Yet these classes are not really meant to be extended 🤷‍♂️  We no longer add the final keyword to help modules extend them if they REALLY want to but does it mean we should block the whole framework for such a rare use case? (if it ever happens)

jolelievre: I'm just worried that if we block service constructors and consider them as BC break then we can't:
- fix bugs
- add features
- optimize them

This means the minor release will likely become empty scopes that can only fix bugs on legacy code

0017-backward-compatibility-promise.md
jolelievre: Ah I didn't see this part before my previous comments but they remain legit nonetheless

jolelievre: About CQRS I would say we shouldn't remove existing getters but maybe we should allow adding some? If we allow constructor to be updated it likely means new data is returned via the query so a getter is also needed

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+|                                                                |                |
+
+#### CQRS
+
+| Type of Change            | Change Allowed |
+|---------------------------|----------------|
+| Queries                   | No             |
```
jolelievre, line 51: I don't understand what this line stands for?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| QueryResult (constructor) | yes            |
+| QueryResult (getters)     | No             |
+|                           |                |
```
jolelievre, line 54: What about commands?

0017-backward-compatibility-promise.md
jolelievre: Then why not?

```suggestion
| Add strict typing      | No             |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| Force strict typing      | No             |
+|                          |                |
+
+#### Symfony Controller actions
+
+Since controllers can be decorated, actions are not in the public API.
+
+#### Experimental Features
+
+Experimental Features and code should be marked with the `@internal` tags.
```
jolelievre, line 73: I agree on this one! Although maybe `@experimental` would be more fitting The internal stuff to me is something that is not supposed to be used directly by external developers (or at their risk) and only internally meaning we can change it the way we want because it's not considered public at all

`@experimental` however is some code that is not finished yet (behind a feature flag for example) so the API is likely to evolve BUT it will ultimately potentially become public

In both cases, it means the API is not considered public and can change, but the intention and goal are a bit different So maybe we could use both tags?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+### Front
+
+#### Back office
+
+| Type of Change                         | Change Allowed |
+|----------------------------------------|----------------|
+| HTML markup                            | Yes            |
+| `id` and `class`                       | No             |
+| Npm dev dependencies                   | Yes            |
+| Npm dependencies                       | No             |
```
jolelievre, line 88: Isn't it allowed if the update doesn't introduce BC breaks? Or do you mean removing a dependency?
What about adding one?

0017-backward-compatibility-promise.md
jolelievre: Move file or change file name is the same to me 🤔

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+
+| Type of Change                                            | Change Allowed |
+|-----------------------------------------------------------|----------------|
+| Change/remove routes                                      | No             |
+| Add routes                                                | Yes            |
+| Javascript Core functions (name, parameters, return type) | No             |
+| Remove injected variables into template engines           | No             |
+| Add variables into template engines                       | Yes            |
+| Change/remove template file name                          | No             |
+| Change/remove javascript file name [2]                    | Yes            |
```
jolelievre, line 105: By the [2] you mean it's allowed to rename sources but not built assets?
What about the generic components?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| Functions directly exposed to the DOM  | No             |
+| Change/remove template file name       | No             |
+| Change/remove javascript file name [2] | Yes            |
+|                                        |                |
+
+#### Front office
+
+| Type of Change                                            | Change Allowed |
+|-----------------------------------------------------------|----------------|
+| Change/remove routes                                      | No             |
```
jolelievre, line 99: What about this one in BO?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+### Webservices
+
+| Type of Change                     | Change Allowed |
+|------------------------------------|----------------|
+| Change/remove routes               | No             |
+| Add routes                         | Yes            |
+| Remove elements in XML/JSON struct | No             |
+| Add elements in XML/JSON struct    | Yes            |
+| Remove parameters                  | No             |
+| Add parameters                     | Yes            |
```
jolelievre, line 134: Maybe only optional? Additional required parameters would be a BC no?
```suggestion
| Add optional parameters                     | Yes            |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Database structure
+
+| Type of Change                              | Change Allowed |
+|---------------------------------------------|----------------|
+| Change/remove field type or name to a table | No             |
+| Change field default value                  | No             |
+| Change/remove table name                    | No             |
+| Add new optional field to a table           | Yes            |
+| Introducting a required filed               | No             |
```
jolelievre, line 146:
```suggestion
| Introducting a required field               | No             |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+| Change/remove constraints                   | Yes            |
+|                                             |                |
+
+
+[1] Not if the class is tagged as `@internal`
```
jolelievre, line 151: I guess we're going to add a bunch of `@internal` tags then 😅

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+[2] Only if it's not a production file
```
jolelievre, line 152: Despite my comments I globally agree with these definitions and the idea of this promise However I wonder if we should moderate this promise with a paragraph explaining that this is the definition of what we consider a breaking change And that we try as hard as possible no to add breaking change in minor versions BUT that some cases force us to do so and details which cases would force us to it:
- you already mentioned security fixes
- maybe critical bugs that can't be fixed without breaking change is acceptable as well
- what about regresions?
- maybe other cases...

jolelievre (commented):
Hmmm this is interesting ^^

I commented most of my worries, the biggest one being the possibility of changing services constructor parameters. That said if we formalize the use of the `@internal` tag and use it extensively (meaning we would need to go through all our services before releasing the v8 to clearly state which are public and which are not) then it might limit my worries

If we do so I would add a breaking change: `Turn "public" code into "internal" one`

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+### PHP
+
+#### General purpose
+
+| Type of Change                                                 | Change Allowed |
+|----------------------------------------------------------------|----------------|
+| Interface/class removal                                        | No             |
+| Public and protected method removal [1]                        | No             |
+| Change/remove service name                                     | No             |
+| Composer dependencies upgrade without BC breaks                        | Yes            |
```
eternoendless, line 33:
```suggestion
| Upgrade composer dependencies which include BC breaks                        | Yes            |
```

0017-backward-compatibility-promise.md
eternoendless: Consider listing them: Guzzle, Doctrine, ...?

0017-backward-compatibility-promise.md
eternoendless: How about we exclude service constructor arguments from the API promise? It makes sense b/c those are classes that you aren't supposed to instantiate yourself. Service overrides are hard to be made forward compatible unless it's done via service decoration, so it would make sense.

0017-backward-compatibility-promise.md
eternoendless:
```suggestion
| Change/remove Query constructor parameters                   | No             |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+
+Since controllers can be decorated, actions are not in the public API.
+
+#### Experimental Features
+
+Experimental Features and code should be marked with the `@internal` tags.
+
+#### Security issues
+
+Note that backward compatibility breaks are tolerated if they are required to fix a security issue.
```
eternoendless, line 77: This is a decision, not an information page, so:
```suggestion
Backward compatibility breaks are tolerated if they are required to fix a security issue.
```

0017-backward-compatibility-promise.md
eternoendless: Usually adding is always allowed, unless when dealing with parameters

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+#### Back office
+
+| Type of Change                         | Change Allowed |
+|----------------------------------------|----------------|
+| HTML markup                            | Yes            |
+| `id` and `class`                       | No             |
+| Npm dev dependencies                   | Yes            |
+| Npm dependencies                       | No             |
+| Functions not exposed to the DOM       | Yes            |
+| Functions directly exposed to the DOM  | No             |
```
eternoendless, line 90: AFAIK DOM is only about the HTML document. So this should be changed to "global scope". I also thought about exported functions, but they're not covered by the API unless they are also available in the global scope.
```suggestion
| Functions not exposed to the global scope       | Yes            |
| Functions directly exposed to the global scope  | No             |
```

0017-backward-compatibility-promise.md
PierreRambaud: The change shouldn't be allowed if it includes BC breaks :thinking:

0017-backward-compatibility-promise.md
PierreRambaud: We can add both tags. But internal means, we do whatever we want with it, experimental or not

0017-backward-compatibility-promise.md
PierreRambaud: I don't know, I think we should be able to do whatever we want with routes in the BO as, it identifies by a route name :thinking:

0017-backward-compatibility-promise.md
eternoendless: You're absolutely right, I updated my suggestion

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+|                           |                |
+
+#### Methods and functions
+
+| Type of Change           | Change Allowed |
+|--------------------------|----------------|
+| Rename                   | No             |
+| Default parameter values | No             |
+| Return type              | No             |
+| Add parameter            | No             |
```
atomiix, line 64: Add optional parameter?

0017-backward-compatibility-promise.md
atomiix: Maybe we should reuse the same logic as for composer: Upgrade npm dependencies that don't produce BC breaks?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+|                                                           |                |
+
+#### Default theme
+
+| Type of Change                        | Change Allowed |
+|---------------------------------------|----------------|
+| HTML markup                           | Yes            |
+| `id` and `class`                      | No             |
+| Npm dev dependencies                  | Yes            |
+| Npm dependencies                      | No             |
```
atomiix, line 117: Same here, maybe only allow updates that don't produce BC Breaks?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+
+### PHP
+
+#### General purpose
+
+| Type of Change                                                 | Change Allowed |
+|----------------------------------------------------------------|----------------|
+| Interface/class removal                                        | No             |
+| Public and protected method removal [1]                        | No             |
+| Change service name                                     | No             |
```
matks, line 32: Maybe we should clarify it's a Symfony service 😋 and not a service like mysql?

0017-backward-compatibility-promise.md
matks: > How about we exclude service constructor arguments from the API promise?

I agree with this. We can always list them for information if this is needed to help module developers

0017-backward-compatibility-promise.md
matks: I agree  to add both `@internal` and `@experimental`

0017-backward-compatibility-promise.md
matks: Even PHP introduces BC breaks inside minor versions.

I agree with @jolelievre : we should state that "This is the backward compatibility promise we aim to keep however we might be forced to stray away from it sometimes. We will provide a clear list of such BC breaks in any Changelog / Core changes doc and everytime we have to break this promise it's to handle a critical matter (security, critical bug, fixing a bug...)"

0017-backward-compatibility-promise.md
matks: In the "general section" @PierreRambaud wrote "Upgrade composer dependencies that don't produce BC breaks"

matks (commented):
Hi @PierreRambaud there is some things confusing between sections 😅  is this because you are right now re-writing the ADR?

0017-backward-compatibility-promise.md
MeKeyCool: I agree, IMO CQRS philosophy asks more attentions to command than queries. Queries can easily be duplicated cause they don't impact data and application as deep as commands.

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+| Npm dev dependencies                  | Yes            |
+| Npm dependencies                      | No             |
+| Functions not exposed to the DOM      | Yes            |
+| Functions directly exposed to the DOM | No             |
+| Change/remove template file name      | No             |
+|                                       |                |
+
+### Directory structure
+
+All directories name and structure are part of the public API until the split between public and private directories is done.
```
MeKeyCool, line 125: For deployment/test purpose it would be really enjoyable to be able to move all the directories used for cache/logs/assets/...
Can't we create a dedicated static class with a consistent and efficient interface to handle directories access ?
This way we could encapsulate this question with this class / interface and don't see filesystem structure as an API ^^'

matks: @MeKeyCool We do not create this backward compatibility just to be annoying 😅 . The problem behind is upgrades.

Imagine today someone installs a shop and puts a logrotate on folder shop/var/logs.
Tomorrow we move var/logs to misc/logs . The sysadmin upgrades to new version of PrestaShop, boom the logrotate is not cleaning the right folder, maybe even worse it might now be rotating wrong files.

So the folder structure goes beyond code ;) it's not because our php code might be able to abstract the filesystem that nothing else is touching it.

matks:
@PierreRambaud Would you like that I continue applying feedbacks on this ADR?

PierreRambaud:
@matks Of course! Do whatever you want with this :)

0017-backward-compatibility-promise.md
MeKeyCool: I think I haven't been clear about my point. :thinking:
Let me try to explain my whole point of view about the subject. This way, you may be able to point out what you think is wrong in my proposal.

## Context

* Today, prestashop comes with lot of "runtime used" files and directories
> {ps_repos}/.htaccess
> {ps_repos}/app/config/parameters.php
> {ps_repos}/app/config/parameters.yml
> {ps_repos}/app/logs
> {ps_repos}/app/Resources/translations
> {ps_repos}/config/settings.inc.php
> {ps_repos}/config/themes/classic
> {ps_repos}/download
> {ps_repos}/img/c
> {ps_repos}/img/genders
> {ps_repos}/img/l
> {ps_repos}/img/m
> {ps_repos}/img/p
> {ps_repos}/img/os
> {ps_repos}/img/st
> {ps_repos}/img/su
> {ps_repos}/img/tmp
> {ps_repos}/log
> {ps_repos}/mails
> {ps_repos}/modules
> {ps_repos}/themes/classic/assets/cache
> {ps_repos}/translations
> {ps_repos}/upload
> {ps_repos}/var/bootstrap.php.cache
> {ps_repos}/var/cache
> {ps_repos}/var/logs
> {ps_repos}/var/sessions
* All those files/directories are necessary for Prestashop to work well.
* All those files/directories implies some attention from administrators (security, meta analysis, control, ...). 
    > :point_up: Having those files/directories "hard coded" can be a pain, especially cause they in the middle of sources mixed with versioned files.


## Goals

We have two different profiles concerned by filesystem structure : 
* **developers** : they want to know where to store/access/link their files, configuration, assets, ... from source code.
* **administrators** : they want to be able to copy / backup / restore all those files and directory contents. Furthermore, they want to be able to manage read/write/edit access rights. With modern deployment tools, some of them may want to be able to customize their location.

> :point_up: Obviously, if administrator change anything in filesystem it should not break application behavior.

* **For developers**, we want to ensure that they can store/access/link their files, configuration, assets, ... in the proper directory without changing their code (if they have to, it would be a BC).
* **For administrators**, we want to ensure that they don't have to change any script/configuration for patch or minor release update.

As we have different profiles with different usages that should interfere on the same stateful object (filesystem), we have to define a contract, an API to ensure they work together without breaking each other work.

And obviously, this document is about changes that could impact both of them.



## What I understand from the proposal "*All directories name and structure are part of the public API until the split between public and private directories is done.*"

For me, solving the problem this way implies that the filesystem structure is the direct API and any change would be a BC.

For me, this solution negate completely the needs of devops. It is like "you want to extract logs to be able to put them anywhere else ?" => **Breaking Change**, "you want to move assets to optimize loadbalancing or static files management ?" => **Breaking change** 


## My point

The idea is just to change API for developers/contributors to a dedicated code. This way, we could create a real devops roadmap without breaking changing API at every change.

It is probably not optimal solution but let me share a simple roadmap example to what I have in mind :

1. Create a dedicated static class used to abstract filesystem.
    > ```php
    > final class PrestashopFilesystem
    >   {
    >       private static function defaultPathMap() {
    >           return [
    >               'app/logs' => 'app/logs',
    >               'var/logs' => 'var/logs',
    >               '...'
    >           ];
    >       }
    >   
    >       public static function getPath(string $pathKey) {
    >           return realpath( _PS_ROOT_DIR_.'/'. self::defaultPathMap()[$pathKey] );
    >       }
    >           
    >   }
    >
    > ```
2. Add some environment variables (such `PS_FS_APP_LOGS_DIR` or `PS_FS_VAR_LOGS_DIR`) that are used by `PrestashopFilesystem` to make filesystem structure configuration possible for devops and safe for developers.
    > :point_up: if you define `PS_FS_APP_LOGS_DIR` default value to `app/logs`, you should not observe any breaking change for old install or for project that don't care about controlling filesystem.

    > :point_up: If necessary, we even can imagine a system that detect any change in `PrestashopFilesystem` configuration and ensure data consistency.
    >
    > If this feature is seen as necessary, I would recommend to separate this feature to a third roadmap step and to use feature/experimental flags until the whole feature is finished.

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Packages
+
+The PrestaShop code base is composed of two different types of packages:
+
+ - **Production packages:** these are packages that are shipped in the final release build
+ - **Development packages:** these are tools that can be used by third-party developers to lint, test, format or build whatever they want. They are used as npm or composer dependencies.
+
+Backward compatibility guarantees only apply to the production packages except for modules which are not part of the public API.
+
```
Progi1984, line 22: Should we allow update of (prod) dependencies in core ?

I think we should all minor updates but not major updates.

0017-backward-compatibility-promise.md
Progi1984:
```suggestion
| Change Symfony service name                                     | No             |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+### PHP
+
+#### General purpose
+
+| Type of Change                                                 | Change Allowed |
+|----------------------------------------------------------------|----------------|
+| Interface/class removal                                        | No             |
+| Public and protected method removal [1]                        | No             |
+| Change service name                                     | No             |
+| Remove service                                          | No             |
```
Progi1984, line 33:
```suggestion
| Remove Symfony service                                          | No             |
```

0017-backward-compatibility-promise.md
Progi1984: Define in general purpose. Sorry.

0017-backward-compatibility-promise.md
Progi1984: I think we should remove this line as there is no action (removal, change, and so on).

Progi1984:
```suggestion
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,154 @@
...
+| Upgrade composer dependencies that don't produce BC breaks                | Yes            |
+| Composer dependencies used a lot                               | No             |
+| Updating/removing composer dev dependencies                                      | Yes            |
+| Change Hook name                                               | No             |
+| Remove hook parameters                                         | No             |
+| Add hook parameters                                            | Yes            |
+| Change value of a constant                                     | Yes            |
+| Remove or rename constants                                     | No             |
+| Adding argument to an event                                    | Yes            |
+| Modifying the types of thrown exceptions                       | No             |
```
Progi1984, line 43:
```suggestion
| Modifying the type of thrown exceptions                       | No             |
```

0017-backward-compatibility-promise.md
Progi1984: `@internal` is referenced but not `@experimental` : https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/index.html#tag-reference

0017-backward-compatibility-promise.md
Progi1984:
```suggestion
| Upgrade npm dependencies that don't produce BC breaks                       | No             |
```

0017-backward-compatibility-promise.md
Progi1984:
```suggestion
| Upgrade npm dependencies that don't produce BC breaks                      | No             |
```

0017-backward-compatibility-promise.md
Progi1984:
```suggestion
| Add required parameters                     | No            |
| Add optional parameters                     | Yes            |
```

Progi1984 (commented):
Some feedbacks

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,152 @@
...
+Date: 2022-03-24
+
+## Status
+
+N/A
+
+## Context
+
+If you are a theme or module developer, the following guidelines will help you to ensure smooth upgrades to all futur minor releases.
+This page describes rules and best practices for backward compatible development.
```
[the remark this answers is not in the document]
matks: FIxed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: @jolelievre I'm not sure to understand both @PierreRambaud intent and your remark 😅

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

0017-backward-compatibility-promise.md
matks: Fixed in https://github.com/PrestaShop/ADR/pull/25/commits/8bfb3edf1d8388a237110fcb4618ecb65868686f

matks:
@eternoendless @Progi1984 @atomiix @jolelievre I updated the ADR please read again

0017-backward-compatibility-promise.md
matks: @MeKeyCool 

> The idea is just to change API for developers/contributors to a dedicated code. This way, we could create a real devops roadmap without breaking changing API at every change.

This is days and weeks of effort and many impactful changes that might break the software. I understand you mean well but "we just need to change API" is not helping. This ADR attempts to define a Backward Compatibility promise that works with the current state of the software, you're not suggesting something that works _today_ but something that would _maybe work tomorrow_.

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,171 @@
...
+| Throw a new type of exception from an existing method               | Yes            |
+| Adding a optional constructor parameter at the end of the args list | Yes            |
+|                                                                     |                |
+
+#### Composer dependencies
+
+| Type of Change                                                              | Change Allowed |
+|-----------------------------------------------------------------------------|----------------|
+| Upgrade composer dependencies that do not introduce BC breaks               | Yes            |
+| Composer dependencies with a strong impact on userland code and modules [2] | No             |
```
Progi1984, line 56: Whats is `strong impact` ? Could you define it, please ?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Database structure
+
+| Type of Change                              | Change Allowed |
+|---------------------------------------------|----------------|
+| Change/remove field type or name to a table | No             |
+| Change field default value                  | No             |
+| Change/remove table name                    | No             |
+| Add new optional field to a table           | Yes            |
+| Introducting a required field               | No             |
```
Progi1984, line 164:
```suggestion
| Add a required field               | No             |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,171 @@
...
+| Composer dependencies with a strong impact on userland code and modules [2] | No             |
+| Updating/removing composer dev dependencies                                 | Yes            |
+
+#### CQRS
+
+PrestaShop uses Commands and Queries. Specific rules apply to these PHP objects that define a contract.
+
+| Type of Change                              | Change Allowed |
+|---------------------------------------------|----------------|
+| Change a QueryResult constructor            | Yes            |
```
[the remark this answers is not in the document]
Progi1984: Miss rules for Commands.

Progi1984 (commented):
There are some feedbacks.

0017-backward-compatibility-promise.md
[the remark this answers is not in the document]
matks: I think there is no specific rules for Commands. Regular PHP BC break contract apply to Commands. So I add nothing.

QueryResult has a specific rule because we expect userland code will never instantiate them but just receive them so getters are protected but constructor is not.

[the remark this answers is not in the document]
matks: You talked about Commands. Now you talk about Queries? I don't follow you.

If you override a QueryHandler you do whatever you want keep QueryResult, use another, do nothing... 🤔 I don't understand your point.

Also I don't think this backward compatibility promise is built with the idea of "what if someone overrides [...]" ? If you look at it it's focused on what we need to preserve and what we can change when working on the Core.

0017-backward-compatibility-promise.md
matks: Do you have a better definition?

Symfony has a strong impact because modules that do use it will be coupled to controller structure, folder structure, Twig...
Doctrine has a strong impact because modules that do use it will be coupled to the ORM, probably use annotations...

"Strong impact" = "changing this dependency is likely to require a huge/complex rewrite of a lot of userland code"

matks (approved):
I think this is a good first version, enough for our current needs. We can update it as we find new usecases.

eternoendless (approved):
I agree with @matks 

I think we can adopt this first version and add (or remove) stuff as needed in the future.

NeOMakinG (approved):

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,173 @@
...
+
+Further reading: [The Pain Points](https://build.prestashop.com/news/prestashop-in-2019-and-beyond-part-2-pain-points/#no-clearly-defined-api)
+
+## Decision
+
+PrestaShop is committed to follow the [SemVer convention](https://semver.org/), which states that backward incompatible changes are only allowed in major versions.
+
+Only exceptional circumstances (e.g. fixing a security issue) can allow introducing backward incompatible changes in minor or patch versions. Whenever this happens, the maintainer team will make sure to provide a clear list of the affected components.
+
+The following lines detail or backward compatibility promise, and describe what changes are allowed or not in minor or patch versions. Further restrictions (or allowances) might be added in the future.
```
jolelievre, line 23:
```suggestion
The following lines detail our backward compatibility promise, and describe what changes are allowed or not in minor or patch versions. Further restrictions (or allowances) might be added in the future. 
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,173 @@
...
+
+Our backward compatibility promise only applies to production packages, except for modules which are not part of the public API.
+
+### PHP
+
+#### General purpose
+
+Note: "Symfony" is mentioned below as relating to the Symfony framework and its components, including the service container.
+
+| Type of Change                                                      | Allowed |
```
jolelievre, line 40:
```suggestion
| Type of Change                                                      | Allowed        |
```

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,173 @@
...
+| Update a production package with a strong impact on userland code and modules [^2] | No             |
+| Update/remove a dev package                                                        | Yes            |
+
+#### CQRS
+
+PrestaShop uses Commands and Queries which are part of the public API. Specific rules apply to these objects.
+
+| Type of Change                              | Allowed |
+|---------------------------------------------|----------------|
+| Change a QueryResult constructor            | Yes            |
```
jolelievre, line 71: What about commands? Is it forbidden by default unless parameters are optional (but it's already stated in the general purposes)

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,173 @@
...
+
+Since classes whose instantiation is delegated to the Symfony service container are not supposed to be instantiated individually, changing the constructor of those classes is allowed.
+
+#### Symfony Controller actions
+
+Since controllers can be decorated, actions are not in the public API.
+
+#### Experimental Features
+
+Experimental Features and code should be marked with the tags `@internal` or `@experimental` .
```
jolelievre, line 94: I think both tags have a different meaning even if they imply the same concept that we don't promise anything:
- `@experimental` is some code still under development that may change until it's officially released
- `@internal` is a released production code, that is technically extendable (no final) because we may need it internally, but is used for internal purposes and can be modified at will

I think we should detail the difference somewhere, probably a dedicated paragraph for internal

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,173 @@
...
+| Change/remove javascript file name [^3]                   | Yes            |
+
+
+#### Front office
+
+| Type of Change                                            | Allowed |
+|-----------------------------------------------------------|----------------|
+| Change/remove routes                                      | No             |
+| Add routes                                                | Yes            |
+| Change Javascript Core functions (name, parameters, return type) | No             |
```
jolelievre, line 122: Even optional parameters at the end of the list?

0017-backward-compatibility-promise.md
```diff
@@ -0,0 +1,173 @@
...
+| Remove parameters                  | No             |
+| Add optional parameters            | Yes            |
+| Add required parameters            | No             |
+
+
+### Database structure
+
+| Type of Change                              | Allowed |
+|---------------------------------------------|----------------|
+| Change/remove field type or name to a table | No             |
```
jolelievre, line 163: It's a nitpick but is it allowed to change `VARCHAR(32)` to `VARCHAR(64)` for example, we change the limitation but the type is still the same right?

kpodemski (approved):

matthieu-rolland (approved):

0017-backward-compatibility-promise.md
eternoendless: I think internal components could (and probably should) be described in a new ADR. Why don't you submit a proposal?

0017-backward-compatibility-promise.md
eternoendless: Changing is different from adding

jolelievre (approved):

matthieu-rolland:
I have a question regarding the form simplification PRs we have...

Is changing the template of a form using `form_widget` instead of rendering it field by field... considered a BC Break even if it doesn't change the HTML in the end (because all classes, wording etc are transcribed in the form type)

I would say no but I'd like some other dev's input here 🤔

jolelievre:
> I have a question regarding the form simplification PRs we have...
> 
> Is changing the template of a form using `form_widget` instead of rendering it field by field... considered a BC Break even if it doesn't change the HTML in the end (because all classes, wording etc are transcribed in the form type)
> 
> I would say no but I'd like some other dev's input here 🤔

Since the layout doesn't change I don't think it's a BC break it's more like refacto: a different way to attain the same result

I also have one last question: so with this ADR we allow, for example, to change a service constructor's parameter when it's built by the Symfony DI. So our policy allows to do it (also we should try to limit such changes as much as possible), but does it mean we don't need to flag the PR as BC break and not document it? Even if the policy allows it it's probably a good thing to document and warn the community.

Similar question regarding the experimental code (tagged with @experimental) this time however I don't think it's relevant to flag PRs or to document such code modification since the code is by essence destined to be modified until it's still considered experimental.

Maybe those two questions could be answered in a dedicated ADR that comes as an appendix for this one? I'm not sure if the BCbreak documentation is really an architecture decision, it's more about the Project management but if it's not defined here I'm not sure where it should be? 🤷‍♂️ If not in this repo then maybe in this one?https://github.com/PrestaShop/open-source

matks:
> I also have one last question: so with this ADR we allow, for example, to change a service constructor's parameter when it's built by the Symfony DI. So our policy allows to do it (also we should try to limit such changes as much as possible), but does it mean we don't need to flag the PR as BC break and not document it? Even if the policy allows it it's probably a good thing to document and warn the community.

I think people do not care about labels on the PR. What they care about is the list of BC breaks in the devdocs. They will not check every single PR for each BC break.

I think we could mark these modifications in a dedicated section on devdocs 8.0.0 BC breaks page, we can call them 'things that look like BC breaks BUT actually they're not' 😆 . The idea is to provide the information and let users choose what to do with it.

matks (approved):

README.md
```diff
@@ -41,6 +41,7 @@ ADR ID | Date       | Discussion           | Title
 0017   | 2022-04-17 | [Pull Request][0017] | Backward compatibility promise                               | 💬 In discussion
 0018   | 2022-05-16 | ~                    | [Horizontal migration][0018]                                 | ✅ Accepted
 
+
```
FabienPapet, line 44: useless change :)

PierreRambaud:
Could you please go a little faster with this one, I need to clean up my GitHub forks :)

matks:
@PierreRambaud Sorry about delay 😓 2 weeks ago we were close to be able to close the topic with a group agreement however [we found some little funny thing](https://build.prestashop.com/news/major-security-vulnerability-on-prestashop-websites/) and it got delayed

PierreRambaud:
I give you one more week, but maybe you can merge it and keep the ADR in discussion.

eternoendless:
We have 6 approvals, I think we agree on the basics, and I see no fundamental dissent in the discussion. Details can be ironed out in documentation, and any further decisions can be recorded in future ADRs. In the spirit of [lazy consensus](https://community.apache.org/committers/lazyConsensus.html), I'm calling this ADR as accepted.

PR #31, opened 2022-12-27
0020 - BC Break & PHP Abandoned Dependency

Progi1984:
| Questions         | Answers
| Description?      | New ADR
| Type?             | improvement

### DIFF
--- a/0020-bc-break-and-abandoned-dependency.md
+++ b/0020-bc-break-and-abandoned-dependency.md
@@ -1,0 +1,23 @@
+# 20. BC Break & PHP Abandoned Dependency
+
+Date: 2022-11-16
+
+## Status
+
+In discussion
+
+## Context
+
+As said in the ADR #17, impact on production package in Composer dependencies are limited :
+- Updates are only possible only if the package has not a strong impact
+- Removal is not allowed.
+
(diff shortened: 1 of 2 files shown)


Progi1984:
Ping @PrestaShop/committers & @PrestaShop/prestashop-core-developers

eternoendless:
IMO this should be allowed if and only if these are true:

1. The package is not a direct dependency (meaning that it's not in our composer.json)
2. It's not a popular dependency (eg. symfony subcomponents)
3. There's a known, important security vulnerability that cannot be patched easily in a temporary fork
4. (if applicable) an alternative package exists

eternoendless:
BTW it's easier to discuss the subject on an issue, then create the decision record

Progi1984:
@eternoendless So what can we do when direct dependency (it's the target of the ADR) is abandonned during a minor version ? 

I will give a example :  SwiftMailer is not maintained since Nov 2021

eternoendless:
> what can we do when direct dependency (it's the target of the ADR) is abandonned during a minor version ?

Unfortunately we need to keep it unless an important security vulnerability is found and it cannot be patched in a fork.

Progi1984:
@eternoendless Thanks for your feedback. I understand it. I will close this ADR.

PR #33, opened 2023-02-23
0021 - Model and repositories - Reading access

jolelievre:
| Questions         | Answers
| Description?      | Define interfaces representing the structure of our model (representation of the data only without advanced domain intelligence). This aims at removing the direct usage of ObjectModel classes in the recent code.
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | ~
| Sponsor company   | ~
| How to test?      | ~

### DIFF
--- a/0021-model-and-repositories.md
+++ b/0021-model-and-repositories.md
@@ -143,9 +143,18 @@ interface CurrencyRepositoryInterface {
 ```
 
 4. DBAL queries VS Object models
-- Repositories (or sub repositories) based on ObjectModel should be in Adapter namespace
-- Repositories based on DBAL should be in PrestaShopBundle
-- If the repo uses both, it should be in PrestaShopBundle for all methods based on DBAL, the adapter using ObjectModel should be injected inside this repository
+
+- In all cases we will have a base implementation in PrestaShopBundle\Model\Repository that is preferably based on DBAL
+  - the DBAL queries are implemented inside this repository and passed into DTOs
+  - you will need to create a DTO in PrestaShopBundle\Model that implement the interface
+
+Specific cases based on legacy:
+  - If the implementation is based on ObjectModel you create an ObjectModelRepository in Adapter/MyEntityDomain/Repository/MyEntityRepository
+    - the ObjectModel must implement the interface
+    - the ObjectModelRepository is passed as a dependency into the PrestaShopBundle\Repository and is used a proxy
+  - If the implementation is based on a Doctrine Entity you create an EntityRepository in PrestashopBundle/Entity/Repository/MyEntityRepository
+    - the Doctrine entity must implement the interface
+    - the Doctrine repository is passed as a dependency into the PrestaShopBundle\Repository and is used a proxy
 
 ```php
 namespace PrestaShopBundle\Repository;
@@ -155,7 +164,7 @@ use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository as Adpate
 class ProductRepository implements ProductRepositoryInterface {
     private $productRepository;
 
-    public __contruct(AdpaterProductRepository $productRepository) {
+    public __contruct(AdapterProductRepository $productRepository) {
         $this->productRepository = $productRepository;
     }
 
--- a/0021-model-and-repositories.md
+++ b/0021-model-and-repositories.md
@@ -78,7 +78,7 @@ will evolve and handle typed array, then we may be able to improve this).
 
 ### Get entity by ID:
 - multi lang entities MUST specify a languageId
-- multi shop entities MUST specify a shopId
+- multi shop entities which content can vary depending on shop MUST specify a shopId (it it's only a basic association without content modification a simple `getAssociatedShopIds()` is enough and the `shopId` parameter is not necessary)
 - multi shop entities which have more complex multi shop behaviour MAY have a getter based on ShopConstraint
 - entities which are neither multi shop nor multi lang don't need these parameters of course
 
@@ -184,7 +184,7 @@ Specific cases based on legacy:
 **DBAL method example**
 
 ```php
-namespace PrestaShop\PrestaShop\Model\Repository;
+namespace PrestaShopBundle\Model\Repository;
 
 class AccessRepository implements AccessRepositoryInterface
 {
@@ -212,7 +212,7 @@ class AccessRepository implements AccessRepositoryInterface
 **ObjectModel example**
 
 ```php
-namespace PrestaShop\PrestaShop\Model\Repository;
+namespace PrestaShopBundle\Model\Repository;
 
 use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository as AdpaterProductRepository;
 
@@ -241,7 +241,7 @@ class ProductRepository implements ProductRepositoryInterface
 **Doctrine example**
 
 ```php
-namespace PrestaShop\PrestaShop\Model\Repository;
+namespace PrestaShopBundle\Model\Repository;
 
 use PrestaShopBundle\Entity\Repository\ShopRepository as DoctrineShopRepository;
 
@@ -282,7 +282,7 @@ class Access
     }
 }
 
-namespace PrestaShop\PrestaShop\Model\Repository;
+namespace PrestaShopBundle\Model\Repository;
 
 use PrestaShop\PrestaShop\Adapter\Security\Access as AdapterAccess;
 
--- a/0021-model-and-repositories.md
+++ b/0021-model-and-repositories.md
@@ -25,6 +25,9 @@ Base folder src/Core/Model (naming tbd)
 - tool configuration a bit more complex but it should be possible with exclude/include rules
 - time spent to maintain the tools
 
+Which rules do we apply with which tools (prevent usage of specific namespaces, keep the namespace clean)
+Use int or VOs
+
 2. Interfaces object/model
 
 Take all classes based on ObjectModel and turn them into an "entity interface" representing each properties and database columns
@@ -140,25 +143,78 @@ interface CurrencyRepositoryInterface {
 ```
 
 4. DBAL queries VS Object models
-5. Which rules do we apply with which tools (prevent usage of specific namespaces, keep the namespace clean)
-6. Where do we put the implementation
-
-7. Use int or VOs
-8. Write operations
-
-### Create
-
-To be decided:
-- we'll need hooks to be triggered provided the created interface
-- should factory/repository trigger hook or be completely hook independent
-- should the hook be dispatched by the handler
-- should we have a hook helper which handles automatical hook dispatch + repository call, this helper (seems like a good separation)
+- Repositories (or sub repositories) based on ObjectModel should be in Adapter namespace
+- Repositories based on DBAL should be in PrestaShopBundle
+- If the repo uses both, it should be in PrestaShopBundle for all methods based on DBAL, the adapter using ObjectModel should be injected inside this repository
 
 ```php
-interface CurrencyFactoryInterface {
-    public function create()
+namespace PrestaShopBundle\Repository;
+
+use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository as AdpaterProductRepository;
+
+class ProductRepository implements ProductRepositoryInterface {
+    private $productRepository;
+
+    public __contruct(AdpaterProductRepository $productRepository) {
+        $this->productRepository = $productRepository;
+    }
+
+    public function getProduct(int $productId, int $languageId, int $shopId): LocalizedProductInterface
+    {
+        return $this->productRepository->getByLanguage(new ProductId($productId), new ShopId($shopId), new LanguageId($languageId));
+    }
+
+    public function getLocalizedProduct(int $productId, int $shopId): LocalizedProductInterface
+    {
+        // Return ObjectModel that has to implement the interface
+        return $this->productRepository->get(new ProductId($productId), new ShopId($shopId));
+    }
 }
 ```
 
-### Update
-### Delete (multishop or not)
+```php
+namespace PrestaShop\PrestaShop\Adapter\Security;
+
+class Access {
+    public function getRoles(): array
+    {
+        $rolesFlattenArray = LegacyAccess::getRoles($employeeProfileId);
+
+        // Transform flatten array into interfaces (implemented by DTO probaly)
+        return array_map(turnArrayToInterface, $rolesFlattenArray);
+    }
+}
+
+namespace PrestaShopBundle\Repository;
+
+use PrestaShop\PrestaShop\Adapter\Security\Access as AdapterAccess;
+
+class AccessRepository implements AccessRepositoryInterface {
+    private $access;
+
+    public __contruct(AdapterAccess $access) {
+        $this->access = $access;
+    }
+
+    /**
+     * @return Role[]
+     */
+    public function getRoles(): array
+    {
+        return $this->access->getRoles();
+    }
+
+    // OR
+
+    /**
+     * @return Role[]
+     */
+    public function getRoles(): array
+    {
+        // Or implement it via DBAL query
+        return $this->connection->createQueryBuilder()
+        ...
+        ;
+    }
+}
+```
--- a/0021-model-and-repositories.md
+++ b/0021-model-and-repositories.md
@@ -51,7 +51,7 @@ interface LocalizedContactInterface {
     public function getEmail(): int;
 
     /**
-     * @return array<int, string> Associative array, the ley is languageId and the value is the localized value.
+     * @return array<int, string> Associative array, the key is languageId and the value is the localized value.
      */
     public function getLocalizedNames(): array;
     public function getLocalizedDescriptions(): array;
@@ -69,11 +69,11 @@ interface LocalizedContactInterface {
 Two repositories are needed for multi lang interfaces since they don't return the same interface. However for multi shop entities the shopId is always mandatory.
 
 ```php
-interface ContactRepository {
+interface ContactRepositoryInterface {
     public function getContact(contactId, languageId, shopId): ContactInterface;
 }
 
-interface LocalizedContactRepository {
+interface LocalizedContactRepositoryInterface {
     /**
      * Returns ALL languages present in the database.
      *
@@ -91,15 +91,74 @@ interface LocalizedContactRepository {
      *
      * @return LocalizedContactInterface
      */
-    public function getLocalizedContactByLanguages(contactId, shopId, languages): LocalizedContactInterface;
+    public function getLocalizedContactWithLanguages(contactId, shopId, languages): LocalizedContactInterface;
+}
+```
+
+When the entity is not found an exception should be thrown `CurrencyNotFoundException` `ContactNotFoundException`
+
+Not found exception MUST extend the associated Domain Exception, but each type of exception MAY implement a generic Exception interface
+
+```php
+class CurrencyNotFoundException extends CurrencyException implements EntityNotFoundExceptionInterface {
 }
 ```
 
 ### Get byXYZ
+
+```php
+interface CurrencyRepositoryInterface {
+    public function getCurrency(currencyId, languageId, shopId): CurrencyInterface;
+    public function getCurrencyByIsoCode(isoCode, languageId, shopId): CurrencyInterface;
+    
+    public function getCurrencyBy(array $criteria, languageId, shopId): CurrencyInterface;
+}
+```
+
+When more than one entity is found (can happen with getByCriteria) an exception should be thrown `NonUniqueCurrencyException` that implements `NonUniqueEntityExceptionInterface`
+
 ### List / search / filter (return IDs or full objects)
-### Create
-### Update
-### Delete (multishop or not)
+
+Decision must be taken between
+- return an array of interfaces (we lose the strong type, but no need to implement a collection for each entity, MAYBE more performant?)
+- return a collection of interfaces
+
+Decision must be taken between
+- findAll
+- getAll
+
+```php
+interface CurrencyRepositoryInterface {
+    /**
+     * @return CurrencyInterface[]
+     */
+    public function findAll(): array;
+    public function getAll(): CurrencyCollectionInterface;
+
+    public function findBy(array $filters = [], array $sortBy = [], ?int $limit = null, ?int $offset = null): array;
+}
+```
 
 4. DBAL queries VS Object models
 5. Which rules do we apply with which tools (prevent usage of specific namespaces, keep the namespace clean)
+6. Where do we put the implementation
+
+7. Use int or VOs
+8. Write operations
+
+### Create
+
+To be decided:
+- we'll need hooks to be triggered provided the created interface
+- should factory/repository trigger hook or be completely hook independent
+- should the hook be dispatched by the handler
+- should we have a hook helper which handles automatical hook dispatch + repository call, this helper (seems like a good separation)
+
+```php
+interface CurrencyFactoryInterface {
+    public function create()
+}
+```
+
+### Update
+### Delete (multishop or not)
--- a/0021-model-and-repositories.md
+++ b/0021-model-and-repositories.md
@@ -1,10 +1,20 @@
 # 21. Model and Repositories aiming to replace the usages of legacy ObjectModel
 
-Date: 2023-02-23
+Date: 2023-10-27
 
 ## Status
 
-In Discussion
+Rejected
+
+## Decision
+
+This proposal will not be applied, the main arguments against it are:
+- the fear of adding a new extra layer that would confuse the community more than it helps it
+- the cost/time it would take to apply this on all the entities in place of other more valuable topics
+- adding this extra layer would indeed hide the legacy core classes, but it doesn't remove their usage, it mostly hides it
+
+Instead of going into that direction it was decided that it would be more valuable for the project to start migrating the core classes to Doctrine entities,
+an ADR has been opened to define the strategy to do this https://github.com/PrestaShop/ADR/pull/37
 
 ## Context
 
--- a/0021-model-and-repositories.md
+++ b/0021-model-and-repositories.md
@@ -1,0 +1,78 @@
+# 21. Model and Repositories aiming to replace the usages of legacy ObjectModel
+
+Date: 2023-02-23
+
+## Status
+
+In Discussion
+
+## Context
+
+Define interfaces representing the structure of our model without advanced domain intelligence. 
+
+### Decisions to make
+
+1. which folder/namespace to use 
+
+Base folder src/PrestaShop (naming tbd)
+- simpler configuration tools and CI
+- no PrestaShop\PrestaShop in namespace
+- harder for contributors to understand this new structure
+- have a new folder with very clean rules, and migrate the Core classes piece by piece into it
+
+Base folder src/Core/Model (naming tbd)
+- follow the structure that has been used so far
+- tool configuration a bit more complex but it should be possible with exclude/include rules
+- time spent to maintain the tools
+
+2. Interfaces object/model
+
+Take all classes based on ObjectModel and turn them into an "entity interface" representing each properties and database columns
+The getters for localized values returns only one field which is already translated
+
+Object with multilang properties have getters for those field which return only string already localized (the language id is specified when fetching the object).
+
+```php
+interface ContactInterface {
+  public function getContactId(): int;
+  public function getEmail(): int;
+
+  // Localized getters
+  public function getName(): string;
+  public function getDescription(): string;
+}
+```
+
+We also have a Localized interface (used mainly in form edition)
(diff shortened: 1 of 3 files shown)


jolelievre:
Note: this PR is a good example of static method replacements that should be anticipated in this ADR https://github.com/PrestaShop/PrestaShop/pull/30271

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+- follow the structure that has been used so far
+- tool configuration a bit more complex but it should be possible with exclude/include rules
+- time spent to maintain the tools
+
+Which rules do we apply with which tools (prevent usage of specific namespaces, keep the namespace clean)
+Use int or VOs
+
+2. Interfaces object/model
+
+Take all classes based on ObjectModel and turn them into an "entity interface" representing each properties and database columns
```
jolelievre, line 33:
```suggestion
Take all classes based on ObjectModel and turn them into an "model interface" representing each properties and database columns
```

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+
+In Discussion
+
+## Context
+
+Define interfaces representing the structure of our model without advanced domain intelligence. 
+
+### Decisions to make
+
+1. which folder/namespace to use
```
jolelievre, line 15: That's the main point that is left to define/decide. Here's my personal suggestion:
| Folder (from root) | Namespace | Content | Rules |
| ----------------- | ------------- | --------| ------ |
| `src/Core/Model` | `PrestaShop/PrestaShop/Core/Model` | All the model interfaces, maybe split by domain at least for large ones like `Product`, should also contain the DTOs if/when needed | Only allows usage of scalar, `DecimalNumber`, `ShopConstraint` and code from the `PrestaShop/PrestaShop/Core/Model` namespace |
| `src/Core/Model/Repository` | `PrestaShop/PrestaShop/Core/Model/Repository` | The repository interfaces (same could be split into sub-namespaces for large domains) | Only allows usage of scalar, `DecimalNumber`, `ShopConstraint` and code from the `PrestaShop/PrestaShop/Core/Model` namespace |
| `src/PrestashopBundle/Repository` | `PrestaShopBundle/Repository` | All the implementations of the interfaces are here, regardless of the implementation (DBAL, ObjectModel, Doctrine, ...) this allows being sure where to find the repository | No usage of legacy objects |
| `src/Adapter/{Model}/Repository` | `PrestaShop\PrestaShop\Adapter\{Model}\Repository` | Repositories based on `AbstractObjectModelRepository` based on `ObjectModel` classes, don't need to implement the model repository interface as it can implement it only partially, method based on static call should also be here (to avoid putting them in the bundle namespace)  | I don't see any particular restriction needed |
| `src/PrestashopBundle/Entity/Repository` | `PrestashopBundle/Entity/Repository` | Repositories based on Doctrine entities | No usage of legacy classes |

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+
+Date: 2023-02-23
+
+## Status
+
+In Discussion
+
+## Context
+
+Define interfaces representing the structure of our model without advanced domain intelligence.
```
eternoendless, line 11: To better understand this proposal:

* What is the exact problem/pain you're intending to solve?
* How does this proposition solve the problem?
* What are the caveats / disadvantages?

eternoendless: This is what the "Context" section should contain

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+    public function getLocalizedNames(): array;
+    public function getLocalizedDescriptions(): array;
+}
+```
+
+3. Repositories
+
+### Get entity by ID:
+- multi lang entities MUST specify a languageId
+- multi shop entities MUST specify a shopId
```
zuk3975, line 68: I think we are missing some specs about those multishop entities that only has associations (shop_id and entity_id, but the entity values are same regardless of shop). So these should not require a shopId in getters as their values are the same for all shops, right?) VS those having different values between shops. It is partially defined in next step about "more complex multishop behavior"

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+
+```php
+interface CurrencyRepositoryInterface {
+    /**
+     * @return CurrencyInterface[]
+     */
+    public function findAll(): array;
+    public function getAll(): CurrencyCollectionInterface;
+
+    public function findBy(array $filters = [], array $sortBy = [], ?int $limit = null, ?int $offset = null): array;
```
zuk3975, line 141: I guess return type depends aswell - either a collection or array (whichever is decided above). Im leaning more towards collection, but only if we manage to create an easily implementable and strongly typed collection

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+use PrestaShop\PrestaShop\Adapter\Product\Repository\ProductRepository as AdpaterProductRepository;
+
+class ProductRepository implements ProductRepositoryInterface {
+    private $productRepository;
+
+    public __contruct(AdapterProductRepository $productRepository) {
+        $this->productRepository = $productRepository;
+    }
+
+    public function getProduct(int $productId, int $languageId, int $shopId): LocalizedProductInterface
```
zuk3975, line 171: As I understand this whole ADR is about `read` models only, where will we put all those `write` methods that currently exists in those Adapter/*/Repository classes? We will still need to call them, but they won't be in the interface (so we'll leave them for now and depend directly from adapter where needed?).

0021-model-and-repositories.md
zuk3975: shouldn't entity & repository interfaces be in somewhere Core/Domain? E.g. `Core/Domain/Product/ProductInterface`| `Core/Domain/Product/Combination/CombinationInterface` | `Core/Domain/Manufacturer/ManufacturerInterface` (maybe wrap in additional /Entity dir as there will already be at least 2 different entity interfaces for one domain), but not sure if its worth wrapping repository, because its one per domain (unless we would keep all the DTO's there as well). Like `Core/Domain/Manufacturer/Repository/ManufacturerRepositoryInterface`, `Core/Domain/Manufacturer/Repository/Result/SomeCustomDTOResultForRepo`

0021-model-and-repositories.md
```diff
@@ -0,0 +1,229 @@
...
+     *
+     * @return LocalizedContactInterface
+     */
+    public function getLocalizedContactWithLanguages(contactId, shopId, languages): LocalizedContactInterface;
+}
+```
+
+When the entity is not found an exception should be thrown `CurrencyNotFoundException` `ContactNotFoundException`
+
+Not found exception MUST extend the associated Domain Exception, but each type of exception MAY implement a generic Exception interface
```
zuk3975, line 103: so we will keep these models|entities and repositories out of domain, but they will throw Domain exceptions? That seems to breaking the hierarchy of exceptions, no? :thinking:

0021-model-and-repositories.md
jolelievre: Good point!!

0021-model-and-repositories.md
jolelievre: For now yes, but a similar ADR is gonna be planned to handle the write operations, that's also when we'll decide if the interfaces should be merged as one or if they will be split

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+    public function getLocalizedContact(int contactId, int shopId): LocalizedContactInterface;
+
+    /**
+     * @param contactId $
+     * @param shopId $
+     * @param languages List of filtered languages
+     *
+     * @return LocalizedContactInterface
+     */
+    public function getLocalizedContactWithLanguages(int contactId, int shopId, int languages): LocalizedContactInterface;
```
eternoendless, line 110: Is this a real use case or a hypothetical one?

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+}
+```
+
+#### Exceptions
+
+**Not found Exception**
+When the entity is not found an exception should be thrown `CurrencyNotFoundException` `ContactNotFoundException`
+
+Not found exception MUST extend the associated Model Exception (`CurrencyException`), but each type of exception MAY implement a
+generic Exception interface.
```
eternoendless, line 120: Technically all exceptions need to extend Exception to be treated as an exception, as explained in [documentation](https://www.php.net/manual/en/class.throwable.php):

> PHP classes cannot implement the Throwable interface directly, and must instead extend [Exception](https://www.php.net/manual/en/class.exception.php).

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+
+(Note we already implemented many DomainException for CQRS, the question here is should we capitalize on them, refactor them? If not
+a mapping Model Exception -> Domain Exception may have to be implemented for existing commands which may be troublesome)
+
+```php
+class CurrencyNotFoundException extends CurrencyException implements EntityNotFoundExceptionInterface {
```
eternoendless, line 126: I think that the exception hierarchy must be thought out, based on expected catch patterns. Are we going to catch exceptions by functional domain (eg Currency), by a infrastructure details (CurrencyStorage / CurrencyRepository) or by general infrastructure (ModelStorage / ModelRepository)?

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+}
+```
+
+**Invalid argument Exception**
+The provided argument should follow some valid values (like being positive ID), when an invalid argument is provided we should throw
+an `InvalidArgumentException` that extends the base `ModelException`.
```
eternoendless, line 132: IMO it would make more sense to respect some logical inheritance chain as mentioned in my comment above. Otherwise we'll have to catch all kinds of different exception types.

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Get byXYZ
+
+```php
+interface CurrencyRepositoryInterface {
+    public function getCurrency(int currencyId, int languageId, int shopId): CurrencyInterface;
+    public function getCurrencyByIsoCode(int isoCode, int languageId, int shopId): CurrencyInterface;
+    
+    public function getCurrencyBy(array $criteria, int languageId, int shopId): CurrencyInterface;
```
eternoendless, line 141: For simplicity's sake, we don't need to repeat Currency:
```suggestion
    public function getById(int currencyId, int languageId, int shopId): CurrencyInterface;
    public function getByIsoCode(int isoCode, int languageId, int shopId): CurrencyInterface;
    
    public function getByCriteria(array $criteria, int languageId, int shopId): CurrencyInterface;
```

eternoendless: Do you already have something in mind on how to handle criteria with an array? In my experience, it's usually worse than a simple SQL string, and you still can't escape parameters using native functions.

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+When more than one entity is found (can happen with getByCriteria) an exception should be thrown `NonUniqueCurrencyException` that extends `CurrencyException`
+and implements `NonUniqueEntityExceptionInterface`
+
+### List / search / filter (return IDs or full objects)
+
+Decision must be taken between
+- return an array of interfaces (we lose the strong type, but no need to implement a collection for each entity, MAYBE more performant?)
+- return a collection of interfaces
```
eternoendless, line 152: IMO it's best not to overengineer the query model
There's doctrine for that!!

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+    public function findAll(): array;
+    public function getAll(): CurrencyCollectionInterface;
+
+    public function findBy(array $filters = [], array $sortBy = [], ?int $limit = null, ?int $offset = null): array;
+}
+```
+
+### 3. DBAL queries VS Object models VS Doctrine entities
+
+- In all cases we will have a base implementation in `PrestaShopBundle\Model\Repository` that is preferably based on DBAL
```
eternoendless, line 172:
```suggestion
- In all cases we will have a base implementation in `PrestaShopBundle\Model\Repository` that is based on DBAL
```
Let's not start something meant to reduce tech debt by using the ancient Db connection, right?

eternoendless: Why would it be in the sf bundle? it should be in the Core

> the Symfony-based subsystem called PrestaShop Bundle is a Symfony bundle that contains Symfony-specific functionality like controllers, forms, views, etc.

(as stated in the [docs](https://devdocs.prestashop-project.org/8/development/architecture/introduction/))

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+  - the DBAL queries are implemented inside this repository and passed into DTOs
+  - you will need to create a DTO in `PrestaShopBundle\Model` that implement the interface
```
eternoendless, line 174: This shouldn't be necessary if use Doctrine entities by default

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+
+Specific cases based on legacy:
+  - If the implementation is based on `ObjectModel` you create an `ObjectModelRepository` in `Adapter/MyEntityDomain/Repository/MyEntityRepository`
+    - the `ObjectModel` must implement the Model interface
+    - the `ObjectModelRepository` is passed as a dependency into the `PrestaShopBundle\Repository` and is used a proxy so it doesn't need to implement the interface itself
```
eternoendless, line 179: So we're going to modify the object models to implement the interface, is that right?

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+### 3. DBAL queries VS Object models VS Doctrine entities
+
+- In all cases we will have a base implementation in `PrestaShopBundle\Model\Repository` that is preferably based on DBAL
+  - the DBAL queries are implemented inside this repository and passed into DTOs
+  - you will need to create a DTO in `PrestaShopBundle\Model` that implement the interface
+
+Specific cases based on legacy:
+  - If the implementation is based on `ObjectModel` you create an `ObjectModelRepository` in `Adapter/MyEntityDomain/Repository/MyEntityRepository`
```
eternoendless, line 177: Have you thought about creating entities for all tables and begin to slowly transform object models into proxies of doctrine entities?

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+    private $shopRepository;
+
+    public __contruct(DoctrineShopRepository $shopRepository) {
+        $this->shopRepository = $shopRepository;
+    }
+
+    public function getShop(int $shopId): ShopInterface
+    {
+        // Return Shop Doctrine entity that has to implement the interface ShopInterface
+        return $this->shopRepository->findById($shopId);
```
eternoendless, line 259: This is a double abstraction that I don't really feel at ease with

app code -> shop repository -> doctrine repo -> SQL

IMO we should access repos directly, and use interfaces for abstraction

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+}
+```
+
+### Localized interface (mostly for CRUD)
+Some use cases require having all the data for all languages (for edition, or multi lang components), then the multilang fields are returned as an associative array
+(language ID used as the key). They don't include the mono-lang getter.
+
+```php
+interface LocalizedContactInterface {
+    public function getContactId(): int;
```
kpodemski, line 57:
```suggestion
    public function getId(): int;
```

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+
+```php
+interface CurrencyRepositoryInterface {
+    /**
+     * @return CurrencyInterface[]
+     */
+    public function findAll(): array;
+    public function getAll(): CurrencyCollectionInterface;
+
+    public function findBy(array $filters = [], array $sortBy = [], ?int $limit = null, ?int $offset = null): array;
```
kpodemski, line 166: I don't feel this `findBy`, here we have a finite number of options to filter, right? why don't we introduce these extra few methods?

0021-model-and-repositories.md
kpodemski: it looks a bit like dependency inception, what are the benefits of this approach @jolelievre ?

0021-model-and-repositories.md
eternoendless: But is there a situation where you want a subset of all languages? My reasoning is that for multilanguage fields, you want ALL languages, not a subset.

0021-model-and-repositories.md
eternoendless, line 166: We should try and avoid rebuilding what Doctrine already does.

0021-model-and-repositories.md
kpodemski: for edit parts in back office you want to have all langs (this is how forms works right now), for front office you want to have a single lang

eternoendless: For a single lang you'd use ContactRepostory, not LocalizedContactRepository

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+  public function getDescription(): string;
+}
+```
+
+### Localized interface (mostly for CRUD)
+Some use cases require having all the data for all languages (for edition, or multi lang components), then the multilang fields are returned as an associative array
+(language ID used as the key). They don't include the mono-lang getter.
+
+```php
+interface LocalizedContactInterface {
```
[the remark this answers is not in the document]
jolelievre: That makes, but we already used the `Localized` prefix in many CQRS commands, isn't there a risk to create inconsistency and confusion?

0021-model-and-repositories.md
jolelievre: The use case is maybe limited but it exists:
- in BO we want to be able to edit all installed languages regardless of their status
- in FO we want only values for enabled languages so we need to filter them

I believe I struggled with this subtility when dealing with CLDR a while back, which forced us to add new methods in interfaces an cause breaking changes to meet all the needs

0021-model-and-repositories.md
jolelievre: Yes, that's not what I meant. All exceptions will extend `Exception` since the model exception itself will extend it `CurrencyException extends CoreException` and `CoreException` already extends `Exception`

But the inheritance only allows "vertical" criteria in the try/catch blocks, I was thinking of maybe having some throwable interfaces to catch exceptions "horizontally" An example will be easier to make me understand, if I want to catch that an entity (any entity) has not been found if all the not found exceptions implement a common (empty) interface `ModelNotFoundInterface` then it's possible to catch it

0021-model-and-repositories.md
jolelievre: So far we often needed to catch by domain `catch (ProductException)` to catch everything related to Product for example, but this mindset may be influenced by the fact we were working on CQRS and so domain exceptions

So you would suggest more generic exceptions like any entity would always throw the same `ModelNotFoundException` for example? What if you want to test that a specific kind of entity has not been found? Maybe we could have common exceptions extend for each model like `CurrencyNotFoundException extends ModelNotFoundException` and `LanguageNotFoundException extends ModelNotFoundException`?

But then if we want to catch any exception related to Currency (not found, invalid argument, later with writing part maybe some constraint exceptions) then we won't have a common exception for Currency only.

0021-model-and-repositories.md
jolelievre: So you think it would be safer not to have a generic `getByCriteria` method, but only explicit ones? I'm fine with that personally, it was just to ease the life of developers a bit. The idea here wasn't to force this method on all repositories, only the ones where it would be needed and relevant. And the form of the criteria we had in mind was a basic associative array based on entity field/value matching (similar to what Doctrine repositories allow)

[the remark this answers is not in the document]
jolelievre: This method would not be mandatory of course, as you see we favored explicit methods like `getById` `getByIsoCode` But we anticipated that SOME repositories (not all of them) may need a more generic method sometimes, in which case implementing an enormous switch case based on criteria just to call explicit methods may not be ideal for developer experience

0021-model-and-repositories.md
jolelievre: I agree, it also makes the implementation rely on more files/classes I personally would prefer rely on simple arrays

0021-model-and-repositories.md
jolelievre: Clearly, this was inspired from doctrine find methods which can be quite convenient But how do we think we could easily search an object then? Because explicit `getByIsoCode` are not search method they are more explicit getters based on a single field, not a search. Or do you think the repositories should propose no search feature at all?

0021-model-and-repositories.md
jolelievre: The purpose was to keep the core namespace free of external classes, especially since they can be based on Doctrine, ObjectModel, DBAL, ... We try to remove legacy/external code from the Core namespace so if our implementation of core repositories already don't respect this it is kind of missing a point and it lacks consistency no?

0021-model-and-repositories.md
jolelievre: If they exist yeah, but not all models have their own Doctrine entity. And the idea of this ADR is not to encourage creating new Doctrine entities but to promote not depending on a fixed solution (that we don't have for now). The DTO is the only one that has the advantage of being independent from the persistence implementation, but if we want to use existing Doctrine entity or existing ObjectModel it would be perfectly acceptable.

0021-model-and-repositories.md
jolelievre: Not really because doctrine is not considered as the sure solution of replacement for ObjectModel as far as I know. If it was this whole ADR wouldn't make sense and we would start migrating to Doctrine already 🤔

0021-model-and-repositories.md
jolelievre: That's the idea yes, to add the getters

0021-model-and-repositories.md
jolelievre: The purpose is to have ALL the model repositories in the same place so there is no debate about where to implement them or find them. Their internal implementation though can be based on multiple solutions depending on the entity. If a Doctrine entity already exists chances are a doctrine repository will be used, if it only has an ObjectModel it would use an ObjectModelRepository, if it's only based on legacy static methods based on Db then it would probably be preferable to implement a DBAL query.

And some model repositories will also probably have a mix of implementations, we already have these needs in recent ObjectModelRepositories that can't handle every queries with ObjectModel only especially "complex" queries with joins that usually. need dedicated DBAL queries

Because of the different solutions, and the possible mix of solutions we though having a single place/folder/namespace for all repositories, even if they are sometimes mear proxies, would bring some consistency. And when they will need refacto the code is also centralized.

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+}
+```
+
+### 4. Namespaces and rules
+
+| Folder (from root) | Namespace | Content | Rules |
+| ----------------- | ------------- | --------| ------ |
+| `src/Core/Model` | `PrestaShop/PrestaShop/Core/Model` | All the model interfaces, maybe split by domain at least for large ones like `Product`, should also contain the DTOs if/when needed | Only allows usage of scalar, `DecimalNumber`, `ShopConstraint` and code from the `PrestaShop/PrestaShop/Core/Model` namespace |
+| `src/Core/Model/Repository` | `PrestaShop/PrestaShop/Core/Model/Repository` | The repository interfaces (same could be split into sub-namespaces for large domains) | Only allows usage of scalar, `DecimalNumber`, `ShopConstraint` and code from the `PrestaShop/PrestaShop/Core/Model` namespace |
+| `src/PrestashopBundle/Repository` | `PrestaShopBundle/Repository` | All the implementations of the interfaces are here, regardless of the implementation (DBAL, ObjectModel, Doctrine, ...) this allows being sure where to find the repository | No usage of legacy objects |
```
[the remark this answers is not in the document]
jolelievre: Hmm, I'm not sure 🤔 To me Core namespace should be about the PrestaShop additional value, its model its domain its business intelligence. Maybe some tools that are specific to PrestaShop.

But the Bundle namespace makes more sense for all that is completely related and dependent with the Symfony framework, like the controllers, event listeners, the form types, ...

0021-model-and-repositories.md
kpodemski: @jolelievre that's surprising, I think we should answer this question, I've seen many times that - eventually, there will be Doctrine entities, I don't recall any other alternative

0021-model-and-repositories.md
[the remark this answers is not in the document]
eternoendless: It's never too late to correct a mistake :)

0021-model-and-repositories.md
```diff
@@ -0,0 +1,315 @@
...
+a code relying on these ObjectModels we create a new technical debt. It may not be a big one, but a technical debt nonetheless since we'll have to
+remove/replace this code at some point.
+
+The purpose of this ADR is to introduce a new access to our data, one that is independent from its actual implementation (ObjectModel, Doctrine, DBAL, ...),
+we propose to add some new interfaces that represent our Model data, it's DATA only these interfaces don't have domain or business logic, they are simply a mean
+to acces our data. We chose interfaces because they have the advantage of being independent of the implementation, it includes the different ways to access the DB,
+but it goes further we could even imagine data accessed on a remote API. With an interface all this internal code doesn't matter any more as long as the interface
+is respected.
+
+**Summary** this ADR proposes to add new **Model interfaces** represent our data model, it also includes some **Repository** interfaces each one being associated to its Model interface.
```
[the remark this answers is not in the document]
eternoendless: Ok so basically you'd make your Doctrine entities implement these interfaces.

0021-model-and-repositories.md
eternoendless: AFAIK in FO you only work with a single language, so you wouldn't need a subset.

0021-model-and-repositories.md
eternoendless: I see. Perhaps that reasoning can be made more clear in the ADR.

0021-model-and-repositories.md
eternoendless: I think it depends on how do you plan to build and catch your exceptions.

As a general rule, I think one should be able to distinguish between domain errors (the Product is not found), integrity errors (the Product exists but is corrupted for some reason), storage or data transport errors (eg the DB is unreachable), and make it so that consumers of Product can choose to handle those errors differently (so you can display different error messages for example) or all at once with a single catch-on-error.

Suppose whenever you build a Product the ProdutRepository has to fetch the related Attributes. Suppose one Attribute doesn't exist although it should. You could make ProductRepository catch AttributeNotFoundException and rethrow it as a ProductException, but that requires ProductRepository to be aware and take care of every single one of their submodels' exception types (eg AttributeException, LanguageException, CurrencyException...) because they don't inherit a common exception type like ModelNotFoundException. But if you do it that way then you can't simply catch ProductException anymore.

It's a tricky dilemma that's been bugging me out for some time, but I don't have a definitive answer. Perhaps marker interfaces are the key.

0021-model-and-repositories.md
eternoendless: I'm worried that we might be adding another abstraction layer on top of Doctrine that will become useless once ObjectModel has been phased out and there's only Doctrine left.

0021-model-and-repositories.md
eternoendless: OK so the interfaces go in the Core and the Doctrine implementation in the bundle?

0021-model-and-repositories.md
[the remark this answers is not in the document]
eternoendless: I'm not completely sure about how good an idea it is to strictly separate business & framework code, given the number of dependencies between the two in our current code.

Eg. Hooks are supposed to become Sf events. When they will be changed, The core would have to create interfaces for the events that would be implemented in the bundle. It can be done for sure, but in the end our interfaces will still be very heavily influenced by the implementation details -- unless we create adapters. Is it smart to pay the price of abstraction upfront, just to be prepared in the off chance we might change the event system?

Now, if we started the software from scratch, we would separate the architecture in four layers:
- presentation (controllers, form types, etc)
- application (CQRS/internal api)
- domain code (rich domain entities, business rules, events)
- infrastructure code (storage/repos)

I feel we're somewhere in between right now. Does our abstraction policy make sense in that perspective?

0021-model-and-repositories.md
eternoendless: We could have specific getters, but delegate the open-ended "find" method to Doctrine

PR #34, opened 2023-05-30
0022 - Module storage folders

tswfi:
| Questions         | Answers
| Description?      | Define new places for modules to save public, private and temporary files.
| Type?             | new feature
| BC breaks?        | no
| Deprecations?     | no
| Fixed ticket?     | 
| Sponsor company   | 
| How to test?      | ?

This is a first draft for the specification for discussion.

Inspired by discussion here: https://prestashop.slack.com/archives/C010UHP5PAR/p1685442917263009

### DIFF
--- a/0022-module_storage_folders.md
+++ b/0022-module_storage_folders.md
@@ -1,0 +1,40 @@
+# 20. Module storage folders
+
+Date: 2023-05-30
+
+## Status
+
+Draft
+
+## Context
+
+Currently module developers save dynamic and static data inside module folder as there is no documented place for
+storing that kind of data.
+
+This causes problems with scaling and makes it harder to keep files "clean" in the module folder.
+
+This ADR defines different places for module developers to save their data.
+
+* Public files
+  * Module api provides a method `getPublicStorageFolder()`
+  * Core prepares new folder for the module in `var/public/modules/[modulename]/`
+  * Module can use that folder as they want (create new subfolders, create files, delete files)
+  * This folder is publicly accessible in `[shopurl]/var/storage/modules/[modulename]`
+* Private files
+  * Module api provides a method `getPrivateStorageFolder()`
+  * Core prepares new folder for the module in `var/private/modules/[modulename]/`
+  * Module can use that folder as they want (create new subfolders, create files, delete files)
+  * This folder IS NOT publicly accessible
+  * Module can still provide a url that can allow the data to be accessed (for example password protected)
+* Temporary files
+  * Module api provides a method `getTmpFolder()`
+  * Core prepares new folder for the module in `var/cache/[environment]]modules/[modulename]/`
+  * Module can use that folder as they want (create new subfolders, create files, delete files)
+  * Core also providers method `getTmpFileName()` that gives a tmpfile name that is unique
+  * This folder is deleted when cache is cleared
+
+### Consequences
+
+Module developers MUST use the api methods to get the folders and keep their files in those folders.
+
+This keeps the module "strictly code" and files are saved elsewhere.


kpodemski:
Thanks @tswfi for submitting this. We'll discuss it with other @PrestaShop/prestashop-maintainers now :)

kpodemski:
https://github.com/PrestaShop/docs/issues/864

0022-module_storage_folders.md
```diff
@@ -0,0 +1,40 @@
...
+
+This causes problems with scaling and makes it harder to keep files "clean" in the module folder.
+
+This ADR defines different places for module developers to save their data.
+
+* Public files
+  * Module api provides a method `getPublicStorageFolder()`
+  * Core prepares new folder for the module in `var/public/modules/[modulename]/`
+  * Module can use that folder as they want (create new subfolders, create files, delete files)
+  * This folder is publicly accessible in `[shopurl]/var/storage/modules/[modulename]`
```
eternoendless, line 22: The path to files should ideally be randomly generated to avoid scraping and/or data leaks (eg to know which module the shop has installed)

0022-module_storage_folders.md
```diff
@@ -0,0 +1,40 @@
...
+Currently module developers save dynamic and static data inside module folder as there is no documented place for
+storing that kind of data.
+
+This causes problems with scaling and makes it harder to keep files "clean" in the module folder.
+
+This ADR defines different places for module developers to save their data.
+
+* Public files
+  * Module api provides a method `getPublicStorageFolder()`
+  * Core prepares new folder for the module in `var/public/modules/[modulename]/`
```
eternoendless, line 20: I think this would be better
```suggestion
  * Core prepares new folder for the module in `var/modules/[modulename]/public/`
```

eternoendless (commented):
May I suggest

* /modules/[modulename] is where the module's CODE goes (and can no longer be written into)
* /var/modules/[modulename] is where modules can READ & WRITE to filesystem
* Inside that folder, you get three subdirectories: `cache`, `private` and `public`
  * `cache` is for, obviously, cached data. Cached data exists for performance reasons and cleared up at any time. The module must be able to reconstruct it at any time.
  * `private` is for private scratch data, like configuration, private keys, non public data, etc. To access this data publicly, the module must implement a dedicated fetch method that validates access to the requested file path.
  * `public` is for public assets, like js, CSS and images. The public path to this folder should not be guessable.

0022-module_storage_folders.md
Hlavtox: I agree with this suggestion.

tswfi: Depends on the public data. If it is static assets like images that are distributed with the module then yes. If it is public data that was uploaded by the merchant/storefront user it should not live inside the module code. Quite a few modules for example create css files according to some settings and those files are specific for that shop.

Hlavtox:
@PrestaShop/committers Hi guys, can we extend on this and possibly push it to finish before v9?

### The propositions before sound good, maybe with a slight change to cache directory:
- `/modules/[modulename]` is where the module CODE goes. Module should not write anything to this folder.
- `/var/cache/[env]/modules/[modulename]` is for cached data. Cached data exists for performance reasons and cleared up at any time. The module must be able to reconstruct it at any time. _This is a slight change to @eternoendless proposition, because I  think it would be much easier to wipe the cache, just by deleting the whole folder._
- `/var/modules/[modulename]` is where modules can READ & WRITE to filesystem to store their data. Inside that folder, you get two subdirectories:
  - `/var/modules/[modulename]/private`  is for private scratch data, like configuration, private keys, non public data, etc. To access this data publicly, the module must implement a dedicated fetch method that validates access to the requested file path.
  - `/var/modules/[modulename]/public` is for public assets, like js, CSS and images.

Question:
- How can access to private directory be blocked? By `.htaccess`? Do we create it dynamically? Maybe it would be easier if the path was `/var/modules/private/[modulename]`, so we can have one main `.htaccess` there and don't have to think about it.

